PR #37, opened 2024-08-23
feat: user auth

choffmann:


choffmann:
I have already implemented the client side login flow in the frontend. When this PR gets merged, it will automatically create a PR in the frontend because the api client has changed. I will rebase my branch to this PR when it is created.

PR #59, opened 2024-09-23
chore: delete not needed fields

doriengr:
I am not sure if this is the right way, but in a create request these fields are calculated by the backend right and not send with the create request. I also do not know if some other places has to be updated. Maybe you have an idea @choffmann ?


choffmann (approved):
:+1: :+1: :+1:

PR #75, opened 2024-09-28
feat: add name to treecluster database

doriengr:
Close green-ecolution/green-ecolution#437

### DIFF
--- a/internal/entities/treecluster.go
+++ b/internal/entities/treecluster.go
@@ -36,5 +36,5 @@ type TreeCluster struct {
 	Longitude      float64
 	Trees          []*Tree
 	SoilCondition  TreeSoilCondition
-	Name           *string
+	Name           string
 }
--- a/internal/server/http/entities/treecluster.go
+++ b/internal/server/http/entities/treecluster.go
@@ -34,7 +34,7 @@ type TreeClusterResponse struct {
 	Longitude      float64                   `json:"longitude,omitempty"`
 	Trees          []*TreeResponse           `json:"trees,omitempty"`
 	SoilCondition  TreeSoilCondition         `json:"soil_condition,omitempty"`
-	Name           *string                   `json:"name,omitempty"`
+	Name           string                    `json:"name,omitempty"`
 } // @Name TreeCluster
 
 type TreeClusterListResponse struct {
@@ -49,7 +49,7 @@ type TreeClusterCreateRequest struct {
 	Description    string                    `json:"description,omitempty"`
 	TreeIDs        []*int32                  `json:"tree_ids,omitempty"`
 	SoilCondition  TreeSoilCondition         `json:"soil_condition,omitempty"`
-	Name           *string                   `json:"name,omitempty"`
+	Name           string                    `json:"name,omitempty"`
 } // @Name TreeClusterCreate
 
 type TreeClusterUpdateRequest struct {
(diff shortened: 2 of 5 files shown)


internal/storage/postgres/migrations/20240928211206_add_name_to_tree_clusters.sql
```diff
@@ -0,0 +1,7 @@
+-- +goose Up
+ALTER TABLE tree_clusters
+ADD COLUMN name TEXT;
```
choffmann, line 3: Can a name be null? If not, add `NOT NULL`

internal/entities/treecluster.go
```diff
@@ -36,4 +36,5 @@ type TreeCluster struct {
 	Longitude      float64
 	Trees          []*Tree
 	SoilCondition  TreeSoilCondition
+	Name           *string
```
choffmann, line 39: I don't think I would use a pointer here

```suggestion
	Name           string
```

internal/server/http/entities/treecluster.go
```diff
@@ -34,6 +34,7 @@ type TreeClusterResponse struct {
 	Longitude      float64                   `json:"longitude,omitempty"`
 	Trees          []*TreeResponse           `json:"trees,omitempty"`
 	SoilCondition  TreeSoilCondition         `json:"soil_condition,omitempty"`
+	Name           *string                   `json:"name,omitempty"`
```
choffmann, line 37:
```suggestion
	Name           string                   `json:"name,omitempty"`
```

internal/server/http/entities/treecluster.go
```diff
@@ -48,6 +49,7 @@ type TreeClusterCreateRequest struct {
 	Description    string                    `json:"description,omitempty"`
 	TreeIDs        []*int32                  `json:"tree_ids,omitempty"`
 	SoilCondition  TreeSoilCondition         `json:"soil_condition,omitempty"`
+	Name           *string                   `json:"name,omitempty"`
```
choffmann, line 52:
```suggestion
	Name           string                   `json:"name,omitempty"`
```

internal/server/http/entities/treecluster.go
```diff
@@ -62,6 +64,7 @@ type TreeClusterUpdateRequest struct {
 	Longitude      float64                   `json:"longitude,omitempty"`
 	TreeIDs        []*int32                  `json:"tree_ids,omitempty"`
 	SoilCondition  TreeSoilCondition         `json:"soil_condition,omitempty"`
+	Name           *string                   `json:"name,omitempty"`
```
choffmann, line 67:
```suggestion
	Name           string                   `json:"name,omitempty"`
```

internal/storage/postgres/treecluster/create.go
```diff
@@ -20,6 +21,7 @@ func defaultTreeCluster() *entities.TreeCluster {
 		Archived:       false,
 		LastWatered:    nil,
 		Trees:          make([]*entities.Tree, 0),
+		Name:           &name,
```
choffmann, line 24: So if you want to change the entities to a non-pointer value, you can use the string here as normal

```suggestion
		Name:           "",
```

Else, when you want to use a Pointer of a string in one line, I have created a utils function 

```go
utils.P("foo") // return &string
```

choffmann (approved):
:fire:

PR #89, opened 2024-10-01
fix: return trees on treecluster create

doriengr:


choffmann (approved):
Yes, good that you found the problem. But I would implement it in the repo. There was already a `WithTrees()` function for the tree cluster repo create function. Now I check and associate the tree cluster id with the trees, right after the tree cluster is created.

PR #102, opened 2024-10-05
feat: implement the delete handler and its logic in the tree service 

### LINKED ISSUE #82: Implement delete tree in tree service

Mohammad96Assaf:
Close #82


internal/server/http/handler/v1/tree/handler.go
```diff
@@ -142,10 +142,18 @@ func UpdateTree(_ service.TreeService) fiber.Handler {
 // @Failure		500	{object}	HTTPError
 // @Router			/v1/tree [delete]
 // @Param			Authorization	header	string	true	"Insert your access token"	default(Bearer <Add access token here>)
```
choffmann, line 144: Please add the id to swagger

```suggestion
// @Router			/v1/tree/{tree_id} [delete]
// @Param			treecluster_id	query	string	false	"Tree Cluster ID"
// @Param			Authorization	header	string	true	"Insert your access token"	default(Bearer <Add access token here>)
```

choffmann (changes requested):
:1st_place_medal:

internal/server/http/handler/v1/tree/handler.go
Mohammad96Assaf: Why treecluster_id, you mean tree_id ?

internal/server/http/handler/v1/tree/handler.go
```diff
@@ -140,12 +140,21 @@ func UpdateTree(_ service.TreeService) fiber.Handler {
 // @Failure		403	{object}	HTTPError
 // @Failure		404	{object}	HTTPError
 // @Failure		500	{object}	HTTPError
-// @Router			/v1/tree [delete]
+// @Router			/v1/tree/{tree_id} [delete]
+// @Param			tree_id			query	string	false	"Tree ID"
```
choffmann, line 144: It's not a query parameter, it's a path parameter.

```suggestion
// @Param			tree_id			path	string	false	"Tree ID"
```

choffmann (approved):
Please change query to path, then you can merge the pr

PR #104, opened 2024-10-08
fix: create and update nil pointer error

choffmann:

Issue #81, opened 2024-09-29
Implement save trees in tree service


choffmann:

choffmann:
Blocked by #68

PR #105, opened 2024-10-08
Feature/implement handler and service logic for saving trees

Mohammad96Assaf:
refs: 
#104 , #81


internal/entities/tree.go
```diff
@@ -19,3 +19,12 @@ type Tree struct {
...
 }
+type TreeCreate struct {
+	TreeClusterID *int32
+	Readonly      bool
+	PlantingYear  int32
+	Species       string
+	Number        string
+	Latitude      float64
+	Longitude     float64
+}
```
choffmann, line 30: Do we really need this?

internal/service/domain/tree/tree.go
```diff
@@ -50,6 +53,42 @@ func handleError(err error) error {
...
+	}
+
+	fn := make([]entities.EntityFunc[entities.Tree], 0)
+	if treeCreate.TreeClusterID != nil {
+		treeClusterID, err := s.treeClusterRepo.GetByID(ctx, *treeCreate.TreeClusterID)
+		if err != nil {
+			return nil, handleError(err)
+		}
+		fn = append(fn, tree.WithTreeCluster(treeClusterID))
+	}
```
choffmann, line 77: If a tree is inserted into an existing tree cluster, the tree cluster coordinates must be updated

choffmann (changes requested):

internal/entities/tree.go
Mohammad96Assaf: For `TreeCluster`, we have both `TreeClusterCreate` and `TreeClusterUpdate` structs for creating and updating, so I followed the same approach for Tree. What do you mean—are you suggesting we use the `TreeCreateRequest` directly in the handler without mapping it to another struct that has the same values?

choffmann: Okay, that's a point... I had a different idea but the more I think about it the smarter it is to do it this way :+1:

choffmann (approved):

Issue #78, opened 2024-09-29
Implement update tree in service


choffmann:
Check if the tree comes from the csv import. For now, only update manuell created trees

choffmann:
Blocked by #68

PR #107, opened 2024-10-09
Feature/ implement handler and service logic for updating trees

Mohammad96Assaf:
refs: 
#78 , #104

 ~~## Note~~
-  ~~Removed `geometry = ST_GeomFromText($10, 4326)` from the query to resolve an issue caused by an inconsistent parameter type in the `UPDATE` query~~


internal/entities/tree.go
```diff
@@ -19,3 +19,13 @@ type Tree struct {
...
+
+type TreeUpdate struct {
+	TreeClusterID *int32
+	Readonly      bool
+	PlantingYear  int32
+	Species       string
+	Number        string
+	Latitude      float64
+	Longitude     float64
+}
```
choffmann, line 31: Do we really need this?

internal/storage/postgres/queries/trees.sql
```diff
@@ -36,8 +36,7 @@ UPDATE trees SET
   latitude = $7,
   longitude = $8,
   readonly = $9,
-  watering_status = $10,
-  geometry = ST_GeomFromText($10, 4326)
+  watering_status = $10
```
choffmann, line 39: That's not the solution to delete `geometry`. It should be like this

```
watering_status = $10,
geometry = ST_GeomFromText($11, 4326)
```

internal/service/domain/tree/tree.go
```diff
@@ -62,3 +66,41 @@ func (s *TreeService) Delete(ctx context.Context, id int32) error {
...
+	}
+	if tu.Species != "" {
+		fn = append(fn, tree.WithSpecies(tu.Species))
+	}
+	if tu.Number != "" {
+		fn = append(fn, tree.WithTreeNumber(tu.Number))
+	}
+	if tu.Latitude != 0 && tu.Longitude != 0 {
+		fn = append(fn, tree.WithLatitude(tu.Latitude), tree.WithLongitude(tu.Longitude))
+	}
```
choffmann, line 86: Yes, okay... another solution, and I think a much cleaner solution, is to use some kind of validator, like https://pkg.go.dev/github.com/go-playground/validator/v10, which I already have somewhere here in this project

internal/service/domain/tree/tree.go
```diff
@@ -62,3 +66,41 @@ func (s *TreeService) Delete(ctx context.Context, id int32) error {
...
+		return nil, handleError(fmt.Errorf("tree with ID %d is readonly and cannot be updated", id))
+	}
+	fn := make([]entities.EntityFunc[entities.Tree], 0)
+	if tu.TreeClusterID != nil {
+		treeCluster, err := s.treeClusterRepo.GetByID(ctx, *tu.TreeClusterID)
+		if err != nil {
+			return nil, handleError(fmt.Errorf("failed to find TreeCluster with ID %d: %w", *tu.TreeClusterID, err))
+		}
+		fn = append(fn, tree.WithTreeCluster(treeCluster))
+	}
```
choffmann, line 74: When a tree gets updated, the old tree cluster coordinate needs to be updated as well

choffmann (changes requested):

internal/service/domain/tree/tree.go
choffmann: see #98

internal/service/domain/tree/tree.go
Mohammad96Assaf: I noticed that you don't use any kind of validation in your implementation for updating or creating tree clusters. I want to ask if we should follow the same approach in the tree cluster section?

internal/service/domain/tree/tree.go
Mohammad96Assaf: Not only the old tree cluster, but I think the new one should also be updated, right?

Mohammad96Assaf (commented):
@choffmann please review the latest edition. I am also waiting for the new component to continue with the functionality for updating the tree cluster coordinates.

internal/entities/tree.go
```diff
@@ -28,3 +28,12 @@ type TreeCreate struct {
...
 	Longitude     float64
 }
+
+type TreeUpdate struct {
+	TreeClusterID *int32  `validate:"omitempty,gt=0"`
+	PlantingYear  int32   `validate:"omitempty,gt=0"`
+	Species       string  `validate:"omitempty"`
+	Number        string  `validate:"omitempty"`
+	Latitude      float64 `validate:"not-zero"`
+	Longitude     float64 `validate:"not-zero"`
```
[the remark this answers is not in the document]
choffmann: and lat and long can both have a zero value

internal/service/domain/tree/tree.go
```diff
@@ -99,5 +105,48 @@ func (s *TreeService) Delete(ctx context.Context, id int32) error {
...
+	if err != nil {
+		return nil, handleError(err)
+	}
+	// TODO: If a new tree cluster has been provided, update the coordinates of both the old tree cluster and the new one.
+	return updatedTree, nil
+}
+func notZero(fl validator.FieldLevel) bool {
+	value := fl.Field().Float()
+	return value != 0
+}
```
choffmann, line 152: Do we really need a custome validator here?

choffmann (changes requested):
:+1:

internal/entities/tree.go
```diff
@@ -28,3 +28,12 @@ type TreeCreate struct {
 	Latitude      float64
 	Longitude     float64
 }
+
+type TreeUpdate struct {
+	TreeClusterID *int32  `validate:"omitempty,gt=0"`
```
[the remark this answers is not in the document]
Mohammad96Assaf: it means that the field is optional and not required for validation. If the field is set to its zero value (the default value for that type, such as 0 for numbers, "" for strings, or nil for pointers), the validator will **skip** further validation (e.g., gt=0 for TreeClusterID) for that field. 


https://pkg.go.dev/github.com/go-playground/validator/v10#hdr-Omit_Empty

choffmann (approved):

PR #168, opened 2024-10-25
Feat: add test cases for import tree

Mohammad96Assaf:
ref: #167


doriengr:
@Mohammad96Assaf Please fix your pipeline 👍

internal/service/domain/tree/import_tree_test.go
```diff
@@ -0,0 +1,193 @@
...
+func TestTreeService_ImportTree(t *testing.T) {
+	ctx := context.Background()
+
+	treeToImport := &entities.TreeImport{
+		Latitude:     54.801539,
+		Longitude:    9.446741,
+		PlantingYear: 2023,
+		Species:      "Oak",
+		Number:       "T001",
+	}
```
doriengr, line 24: Just some code styling: empty line is missing.

internal/service/domain/tree/import_tree_test.go
```diff
@@ -0,0 +1,193 @@
...
+		svc := NewTreeService(treeRepo, sensorRepo, imageRepo, treeClusterRepo, locator)
+
+		// Define existing tree and tree import data
+		existingTree := getTestTrees()[0]
+		treeToImport.PlantingYear = 2024
+
+		// Mock GetByCoordinates to return an existing tree
+		treeRepo.EXPECT().GetByCoordinates(ctx, treeToImport.Latitude, treeToImport.Longitude).Return(existingTree, nil)
+
+		//Mock GetByID because it is called in delete function
```
doriengr, line 112: Do we really need those comments? I think the code is self-explanatory enough 👍 We just need // given, / // when, // then

internal/service/domain/tree/import_tree_test.go
```diff
@@ -0,0 +1,193 @@
...
+			mock.Anything,
+			mock.Anything,
+			mock.Anything,
+			mock.Anything).Return(updatedTree, nil)
+
+		// When
+		err := svc.ImportTree(ctx, []*entities.TreeImport{treeToImport})
+
+		// Then
+		assert.NoError(t, err)
```
doriengr, line 93: Here you have to check if the description is really updated.

internal/service/domain/tree/import_tree_test.go
```diff
@@ -0,0 +1,193 @@
...
+			mock.Anything).Return(expectedTree, nil)
+
+		// When
+		err := svc.ImportTree(ctx, []*entities.TreeImport{treeToImport})
+
+		// Then
+		assert.NoError(t, err)
+	})
+
+	t.Run("should update an existing tree when matching coordinates and planting year are found", func(t *testing.T) {
```
doriengr, line 58: Maybe we can add a test case if the updating fails?

internal/service/domain/tree/import_tree_test.go
```diff
@@ -0,0 +1,193 @@
...
+	ctx := context.Background()
+
+	treeToImport := &entities.TreeImport{
+		Latitude:     54.801539,
+		Longitude:    9.446741,
+		PlantingYear: 2023,
+		Species:      "Oak",
+		Number:       "T001",
+	}
+	t.Run("should create a new tree when no matching tree exists", func(t *testing.T) {
```
doriengr, line 25: Maybe we can add a test case if the creating fails?

internal/service/domain/tree/tree.go
```diff
@@ -106,10 +105,11 @@ func (s *TreeService) Create(ctx context.Context, treeCreate *entities.TreeCreat
 		return nil, handleError(err)
 	}
 
-	if err = s.locator.UpdateCluster(ctx, treeCreate.TreeClusterID); err != nil {
-		return nil, handleError(err)
+	if treeCreate.TreeClusterID != nil {
```
doriengr, line 108: It should be possible to create a tree without a treecluster!

doriengr (changes requested):

internal/service/domain/tree/import_tree_test.go
Mohammad96Assaf: I think we don't need to test the failure cases for update because we have already covered them in tree_test.go.

internal/service/domain/tree/import_tree_test.go
Mohammad96Assaf: I think we don't need to test the failure cases for creation because we have already covered them in `tree_test.go`.

internal/service/domain/tree/import_tree_test.go
Mohammad96Assaf: I don't think we need to check if the update process works correctly, because we've already tested that in `tree_test.go`. It's sufficient to verify that no errors occur after the update. :)

internal/service/domain/tree/tree.go
Mohammad96Assaf: Yes, where is the problem? :)

Mohammad96Assaf:
@doriengr please approve the PR if you don't have any other improvement suggestions. :)

doriengr (approved):

Issue #40, opened 2024-09-08
Implement repository tests


choffmann:
- [x] flowerbed (onhold)
- [x] vehicle
- [x] image (onhold)
- [x] #180
- [x] #179 
- [x] #178
- [x] #177

doriengr:
Finally this epic can be closed 🥳

PR #181, opened 2024-10-28
Setup repository test suite and test region repo

### LINKED ISSUE #180: Implement region repository test

choffmann:
Ref #40 
Close #180


internal/storage/postgres/region/get_test.go
```diff
@@ -0,0 +1,328 @@
...
+import (
+	"context"
+	"testing"
+
+	"github.com/green-ecolution/green-ecolution-backend/internal/entities"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestRegionRepository_GetAll(t *testing.T) {
+	t.Run("Get all regions should return all regions ordered by id", func(t *testing.T) {
```
doriengr, line 12: We should use unified naming of tests. 

I always did something like:
`should return all regions ordered by id`,
`should return empty region list on empty database` …

I do not have a preference what is better.

doriengr (approved):
🔥

PR #184, opened 2024-10-31
feat: create service test

doriengr:


internal/service/domain/service_test.go
```diff
@@ -0,0 +1,46 @@
...
+	"testing"
+
+	"github.com/green-ecolution/green-ecolution-backend/config"
+	"github.com/green-ecolution/green-ecolution-backend/internal/service"
+	"github.com/green-ecolution/green-ecolution-backend/internal/storage"
+	storageMock "github.com/green-ecolution/green-ecolution-backend/internal/storage/_mock"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestNewService(t *testing.T) {
```
Zoka-tech, line 13: At this point I'm not familiar enough with go. But shouldn't it say `t.Run("should return xy", func(t *testing.T)` here like in the other tests?

Zoka-tech (approved):

Issue #130, opened 2024-10-16
Implement service tests


doriengr:
- [x] auth
- [x] info
- [x] region
- [x] tree #139
- [x] treecluster #133 
- [x] sensor

PR #191, opened 2024-11-01
feat: add test for sensor service

### LINKED ISSUE #215: Implement test for sensor service

doriengr:
Ref #130 
Close #215


internal/service/domain/sensor/sensor_test.go
```diff
@@ -0,0 +1,66 @@
+package sensor
+
+import (
+	"context"
+	"testing"
+
+	sensorUtils "github.com/green-ecolution/green-ecolution-backend/internal/server/http/handler/v1/sensor"
```
choffmann, line 7: Maybe this should be a separate package `sensorUtils` or `sensortest` (like `httptest` in standard lib)?

doriengr: We have those files also in the other folders. Maybe we can store the test data all in one file? I don't know what is the possible best way to handle those.

doriengr: 👍

internal/storage/postgres/sensor/get_test.go
```diff
@@ -0,0 +1,389 @@
...
+		// when
+		got, err := r.GetSensorDataByID(ctx, 1)
+
+		// then
+		assert.Error(t, err)
+		assert.Nil(t, got)
+	})
+}
+
+func getTestSensors() []*entities.Sensor {
```
Zoka-tech, line 325: Wouldnt it be better to have the "fake" sensor data in a seperate file like with "internal/service/domain/tree/utils.go" for example?

doriengr:
I am closing this PR. Due to the wrong rebase of mine it is not reviewable anymore. The sensor service tests is also implemented in #209

Issue #194, opened 2024-11-02
Plugin System Backend


choffmann:
- [ ] implement base architecture for the plugin system
- [ ] move csv import to a plugin component

PR #196, opened 2024-11-02
feat: plugin base architecture

### LINKED ISSUE #195: Implement base architecture for the plugin system

choffmann:
Ref #194
Close #195


doriengr (approved):
What about some tests for the new files? Like the Manager for example

choffmann:
> What about some tests for the new files? Like the Manager for example

Will be implementet in #198

PR #207, opened 2024-11-06
Feat: add tests for tree repository

Mohammad96Assaf:


doriengr:
Please use `Close` and not `ref` if you want to close an issue with your PR 👍

internal/storage/postgres/tree/create.go
```diff
@@ -39,16 +39,28 @@ func (r *TreeRepository) Create(ctx context.Context, tFn ...entities.EntityFunc[
...
 	if err != nil {
 		return nil, err
 	}
 
-	if err := r.handleImages(ctx, entity.ID, entity.Images); err != nil {
-		return nil, err
+	entity := defaultTree()
+	for _, fn := range tFn {
+		fn(&entity)
+	}
```
doriengr, line 50: Here is missing a validation. We need to check if specific fields are given in the request otherwise it should return an error. 

For example: For a tree is the longitude and the latitude required. We shouldnot use the demo data for this. This also has to be tested in a new case.

internal/storage/postgres/tree/create.go
```diff
@@ -39,16 +39,28 @@ func (r *TreeRepository) Create(ctx context.Context, tFn ...entities.EntityFunc[
...
+
+	if entity.Images != nil {
```
doriengr, line 52: Why is this needed?

internal/storage/postgres/tree/create_test.go
```diff
@@ -0,0 +1,131 @@
...
+
+		// then
+		assert.NoError(t, err)
+		assert.NotNil(t, got)
+		assert.Equal(t, "", got.Species)
+		assert.NotZero(t, got.ID)
+		assert.WithinDuration(t, got.CreatedAt, time.Now(), time.Second)
+		assert.WithinDuration(t, got.UpdatedAt, time.Now(), time.Second)
+		assert.Nil(t, got.TreeCluster)
+		assert.Nil(t, got.Sensor)
```
doriengr, line 29: Images and Readonly are missing.

internal/storage/postgres/tree/get_test.go
```diff
@@ -0,0 +1,534 @@
...
+			t.Fatal(errGeos)
+		}
+
+		// when
+		lat, long, errRepo := r.GetCenterPoint(context.Background(), ids)
+
+		// then
+		assert.NoError(t, errRepo)
+		assert.Equal(t, g.X(), lat, "Latitude does not match")
+		assert.Equal(t, g.Y(), long, "Longitude does not match")
```
doriengr, line 173: Why is `"Longitude does not match"` returned on successful?

internal/storage/postgres/tree/get_test.go
```diff
@@ -0,0 +1,534 @@
...
+		lat, long, err := r.GetCenterPoint(context.Background(), []int32{})
+
+		// then
+		assert.Error(t, err)
+		assert.Equal(t, 0.0, lat)
+		assert.Equal(t, 0.0, long)
+	})
+}
+
+func TestTreeRepository_GetByTreeClusterID(t *testing.T) {
```
doriengr, line 192: Test cases:
- invalid treeclusterid 0
- invalid treeclusterid -1

internal/storage/postgres/tree/get_test.go
```diff
@@ -0,0 +1,534 @@
...
+		// when
+		trees, err := r.GetByTreeClusterID(ctx, 1)
+
+		// then
+		assert.Error(t, err)
+		assert.Nil(t, trees)
+	})
+}
+
+func TestTreeRepository_GetByCoordinates(t *testing.T) {
```
doriengr, line 242: What about test case with negative latitude/longitude?

internal/storage/postgres/tree/get_test.go
```diff
@@ -0,0 +1,534 @@
...
+		// when
+		images, err := r.GetAllImagesByID(context.Background(), treeID)
+
+		// then
+		assert.NoError(t, err)
+		assert.NotNil(t, images)
+		assert.Empty(t, images, "Images list should be empty")
+	})
+
+	t.Run("should return an empty list when tree is not found", func(t *testing.T) {
```
doriengr, line 327: I think this should return an error and not be successful.

internal/storage/postgres/tree/get_test.go
```diff
@@ -0,0 +1,534 @@
...
+		// when
+		images, err := r.GetAllImagesByID(ctx, 1)
+
+		// then
+		assert.Error(t, err)
+		assert.Nil(t, images)
+	})
+}
+
+func TestTreeRepository_GetSensorByTreeID(t *testing.T) {
```
doriengr, line 357: Maybe here also check for id on 0 and -1

internal/storage/postgres/tree/get_test.go
```diff
@@ -0,0 +1,534 @@
...
+		// when
+		sensor, err := r.GetSensorByTreeID(ctx, 1)
+
+		// then
+		assert.Error(t, err)
+		assert.Nil(t, sensor)
+	})
+}
+
+func TestTreeRepository_GetTreeClusterByTreeID(t *testing.T) {
```
doriengr, line 417: Maybe here also check for id on 0 and -1

internal/storage/postgres/tree/trees_test.go
```diff
@@ -0,0 +1,225 @@
...
+		&generated.InternalTreeRepoMapperImpl{},
+		&generated.InternalImageRepoMapperImpl{},
+		&generated.InternalSensorRepoMapperImpl{},
+		&generated.InternalTreeClusterRepoMapperImpl{},
+	)
+	defer suite.Terminate(ctx)
+	code = m.Run()
+}
+
+func TestTreeRepository_Delete(t *testing.T) {
```
doriengr, line 32: - check for invalid id - 0 and -1
- also check if it is working if no images are set

internal/storage/postgres/tree/trees_test.go
```diff
@@ -0,0 +1,225 @@
...
+		err = r.UnlinkImage(context.Background(), treeID, imageID)
+
+		// then
+		assert.NoError(t, err)
+		isContained, err = isContainedInTreeImages(imageID, treeID)
+		assert.NoError(t, err)
+		assert.False(t, isContained)
+	})
+
+	t.Run("should return error if image or tree not found", func(t *testing.T) {
```
doriengr, line 155: This should be tested in two cases.

internal/storage/postgres/tree/trees_test.go
```diff
@@ -0,0 +1,225 @@
...
+
+		// when
+		err := r.Delete(ctx, 1)
+
+		// then
+		assert.Error(t, err)
+	})
+}
+
+func TestTreeRepository_DeleteAndUnlinkImages(t *testing.T) {
```
doriengr, line 88: - check for invalid id - 0 and -1
- check if error on not found.

internal/storage/postgres/tree/trees_test.go
```diff
@@ -0,0 +1,225 @@
...
+
+		// when
+		err := r.UnlinkImage(context.Background(), 99, 99)
+
+		// then
+		assert.Error(t, err)
+	})
+}
+
+func TestTreeRepository_UnlinkTreeClusterID(t *testing.T) {
```
doriengr, line 168: - check for invalid id - 0 and -1
- check if error on not found.
- check for not set treecluster

doriengr (changes requested):

internal/storage/postgres/tree/create.go
Mohammad96Assaf: I think it would be better to ask @choffmann about this topic, as he is the one who implemented the function. 

However, in my opinion, we don't need to validate the tree attributes again in the repo layer because they are already validated in the service layer. Adding validation in the repo would create redundant checks. 

please have a look on the create function in service tree: 

```
func (s *TreeService) Create(ctx context.Context, treeCreate *entities.TreeCreate) (*entities.Tree, error) {
    if err := s.validator.Struct(treeCreate); err != nil {
        return nil, service.NewError(service.BadRequest, errors.Wrap(err, "validation error").Error())
    }
```
the role of the repo is to handle database interactions while the service layer is reponsible for business logic and validations.

internal/storage/postgres/tree/create.go
Mohammad96Assaf: Please clarify what you mean by "this".

If you are referring to the check entity.Images != nil, it's because the default struct initializes Images as nil. Imagine the function is called and Images is nil—what would happen in that case?

internal/storage/postgres/tree/create_test.go
Mohammad96Assaf: I don't think it's necessary to test every attribute. the aim is to show that the function works as expected. And the handling of images isn't part of the create function, so please take a look at the create function to verify this.

internal/storage/postgres/tree/create_test.go
```diff
@@ -0,0 +1,131 @@
...
+
+import (
+	"context"
+	"github.com/green-ecolution/green-ecolution-backend/internal/entities"
+	"github.com/stretchr/testify/assert"
+	"testing"
+	"time"
+)
+
+func TestTreeRepository_Create(t *testing.T) {
```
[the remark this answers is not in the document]
Mohammad96Assaf: 1. check validation rules
2. check if an error is returned as soon as no longitude or latitude is set
3. I am not sure but maybe the number should also not be nullable (?) that has to be verified
4. check if an error is returned like general speaking as soon as something went wrong for example the request is invalid.

These points are part of the **business logic** (service layer).

internal/storage/postgres/tree/create_test.go
```diff
@@ -0,0 +1,131 @@
...
+		got, err := r.Create(ctx, WithSpecies("Oak"))
+
+		// then
+		assert.Error(t, err)
+		assert.Nil(t, got)
+	})
+}
+
+func TestTreeRepository_CreateAndLinkImages(t *testing.T) {
+	t.Run("should create tree and link images successfully", func(t *testing.T) {
```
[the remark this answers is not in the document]
Mohammad96Assaf: If you look at the create.go file, you'll see that the create function is called initially within it.

1. "Context is canceled": I have already covered this scenario in the test for the create function, so it's not necessary to check it again here.

2. "When there are no images": we test if we can create a tree and link images on it. 
when we don't  give images, then nothing gets linked, which is expected. I don't think testing this is needed, but I'll add it just to be safe.

internal/storage/postgres/tree/create_test.go
```diff
@@ -0,0 +1,131 @@
...
+			WithPlantingYear(2023),
+			WithDescription("Test tree with images"),
+			WithImages(images),
+		)
+
+		// then
+		assert.NoError(t, createErr)
+		assert.NotNil(t, tree)
+		assert.Equal(t, "Oak", tree.Species)
+		assert.Equal(t, "T001", tree.Number)
```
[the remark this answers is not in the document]
Mohammad96Assaf: to ensure the values of the attributes are part of the main create_test. However, I'll add it to be extra safe.

internal/storage/postgres/tree/get_test.go
```diff
@@ -0,0 +1,534 @@
...
+		suite.InsertSeed(t, "internal/storage/postgres/seed/test/tree")
+		r := NewTreeRepository(suite.Store, mappers)
+
+		// when
+		tree, err := r.GetByID(context.Background(), 1)
+
+		// then
+		assert.NoError(t, err)
+		assert.NotNil(t, tree)
+		assert.NotNil(t, tree.TreeCluster)
```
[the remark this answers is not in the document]
Mohammad96Assaf: We reset the database using **suite.ResetDB(t)** and provide the ID **1**. We know exactly how the **test tree** is defined in **02_trees.sql**, and we compare the other attributes to confirm that the retuned tree is the correct one we expect from the get function.
Can you please explain why checking NotNit()  is not enough in this context?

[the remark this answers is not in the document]
Mohammad96Assaf: But We can add it to be extra safe

internal/storage/postgres/tree/get_test.go
```diff
@@ -0,0 +1,534 @@
...
+
+		// when
+		trees, err := r.GetAll(ctx)
+
+		// then
+		assert.Error(t, err)
+		assert.Nil(t, trees)
+	})
+}
+func TestTreeRepository_GetByID(t *testing.T) {
```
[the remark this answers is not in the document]
Mohammad96Assaf: they are not missing, this is covered in the test case "should return error if tree not found": if the id not found then sql will return an error and it is not a matter if the id is -1, 0 or 9999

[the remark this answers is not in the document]
Mohammad96Assaf: or could you explain why I should implement these test cases?

internal/storage/postgres/tree/get_test.go
Mohammad96Assaf: this is the error message. 
Please take a look at this link, it might be helpful: https://pkg.go.dev/github.com/stretchr/testify/assert

internal/storage/postgres/tree/get_test.go
Mohammad96Assaf: I don't understand your question. What difference does it make if the latitude and longitude values are negative or positive? They are just numbers stored in the database.

Please have a look at the file  _**internal\entities\tree.go**_  to see the range validation of the latitude and longitude attributes. Checking these ranges is a part of the service layer.

internal/storage/postgres/tree/create.go
doriengr: You can also take a look here: [https://github.com/green-ecolution/green-ecolution-backend/blob/ec2668316ad13a93c59bc12674cf78aa44270515/internal/storage/postgres/treecluster/create.go#L35](https://github.com/green-ecolution/green-ecolution-backend/blob/ec2668316ad13a93c59bc12674cf78aa44270515/internal/storage/postgres/treecluster/create.go#L35)

This is the treecluster repository which also Cedrik wrote. He is validating everything again. We should unify this!

internal/storage/postgres/tree/create_test.go
doriengr: No the aim is to check, if every data is also returned correctly. It is needed!

internal/storage/postgres/tree/create.go
xk0niR: has been discussed.

internal/storage/postgres/tree/create_test.go
xk0niR: as discussed.

internal/storage/postgres/tree/create_test.go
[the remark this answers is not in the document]
xk0niR: as described.

Mohammad96Assaf:
@doriengr @xk0niR  Please don't forget to review the update and update_test files. Thank you!

doriengr:
The rebase was not done correctly, so this MR is now incredibly difficult to review. We should talk tomorrow about how we want to deal with this …

doriengr:
This PR is closed and reopend in #234

PR #211, opened 2024-11-09
Feature/add router functions to sensors

doriengr:
- [x] GetAll
- [x] GetByID
- [ ] GetDataByID (service function is missing. I'll create a new issue for this)
- [x] Delete

Data is currently not returned Issue https://github.com/green-ecolution/green-ecolution-backend/issues/192

### DIFF
--- a/internal/storage/postgres/sensor/get.go
+++ b/internal/storage/postgres/sensor/get.go
@@ -13,7 +13,7 @@ import (
 func (r *SensorRepository) GetAll(ctx context.Context) ([]*entities.Sensor, error) {
 	rows, err := r.store.GetAllSensors(ctx)
 	if err != nil {
-		return nil, err
+		return nil, r.store.HandleError(err)
 	}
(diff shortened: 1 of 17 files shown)


internal/storage/postgres/sensor/get.go
```diff
@@ -13,7 +13,7 @@ import (
 func (r *SensorRepository) GetAll(ctx context.Context) ([]*entities.Sensor, error) {
 	rows, err := r.store.GetAllSensors(ctx)
 	if err != nil {
-		return nil, err
+		return nil, r.store.HandleError(err)
```
choffmann, line 16: One thing I've learnt in Go so far is that it's always annoying to handle errors in one place. I would just return the error here and the person calling this function takes care of it. Maybe with `errors.Wrap()`. I'm also not happy with the error handling in the service layer...

choffmann (approved):

PR #218, opened 2024-11-10
Feature/add vehicle status and vehicle type

### LINKED ISSUE #217: Implement vehicle service and server

doriengr:
Close #217 

image


choffmann:
Perhaps the return code `400 Bad Request` would be better in this scenario 
image

choffmann:
Also here a `404 not found` code

image

choffmann:
Delete does not work because there are contrains in the db which might be deleted first

image

internal/storage/postgres/mapper/vehicle_test.go
```diff
@@ -95,5 +103,44 @@ var allTestVehicles = []*sqlc.Vehicle{
...
+		expected entities.VehicleStatus
+	}{
+		{input: sqlc.VehicleStatusActive, expected: entities.VehicleStatusActive},
+		{input: sqlc.VehicleStatusAvailable, expected: entities.VehicleStatusAvailable},
+		{input: sqlc.VehicleStatusNotavailable, expected: entities.VehicleStatusNotAvailable},
+		{input: sqlc.VehicleStatusUnknown, expected: entities.VehicleStatusUnknown},
+	}
+
+	for _, test := range tests {
+		t.Run(string(test.input), func(t *testing.T) {
```
Zoka-tech, line 123: To have all tests the same, I think it would be better to have here t.Run("should return etc"

internal/service/domain/vehicle/vehicle_test.go
```diff
@@ -0,0 +1,386 @@
...
+
+		// then
+		assert.False(t, ready)
+	})
+}
+
+func getTestVehicles() []*entities.Vehicle {
+	now := time.Now()
+
+	return []*entities.Vehicle{
```
Zoka-tech, line 364: Maybe have the "fake" data in a seperate file to have this more organized?

doriengr: I only use this fake data in this file.

choffmann (approved):
Approved with changes :fire:

PR #224, opened 2024-11-12
test: utilities

### LINKED ISSUE #223: Test utils

choffmann:
Close #223

### DIFF
--- a/internal/utils/basic_types_mapper.go
+++ b/internal/utils/basic_types_mapper.go
@@ -38,6 +38,10 @@ func StringToNetIP(s string) net.IP {
 }
 
 func NetIPToString(ip net.IP) string {
+  if ip == nil {
+    return ""
+  }
+
 	return ip.String()
 }
 
@@ -51,6 +55,9 @@ func TimeToString(t time.Time) string {
 }
 
 func NetURLToString(u *url.URL) string {
+  if u == nil {
+    return ""
+  }
 	return u.String()
 }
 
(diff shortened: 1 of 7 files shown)


choffmann:
@doriengr there is a function `ParseJSONResponse` in `internal/utils/utils.go` which is only used by tests in the handler. Have you implemented this function? I would like to move this function to the utils files for the http handler. Also, I don't know exactly how to test this function, and if it is only used by tests anyway, we don't need to test the test function.

doriengr:
> @doriengr there is a function `ParseJSONResponse` in `internal/utils/utils.go` which is only used by tests in the handler. Have you implemented this function? I would like to move this function to the utils files for the http handler. Also, I don't know exactly how to test this function, and if it is only used by tests anyway, we don't need to test the test function.

Yes, I created it and as far as I remember I only use it in tests. Feel free to do whatever you want with it 👍

doriengr (approved):
💯

choffmann:
> > @doriengr there is a function `ParseJSONResponse` in `internal/utils/utils.go` which is only used by tests in the handler. Have you implemented this function? I would like to move this function to the utils files for the http handler. Also, I don't know exactly how to test this function, and if it is only used by tests anyway, we don't need to test the test function.
> 
> Yes, I created it and as far as I remember I only use it in tests. Feel free to do whatever you want with it 👍

I wrote a test for this, perhaps it makes sense to use this function outside the tests as well.

PR #234, opened 2024-11-17
Feature/tests for repo tree

Mohammad96Assaf:
close: https://github.com/green-ecolution/green-ecolution-backend/issues/178


doriengr:
The create is not working …
image

internal/storage/postgres/tree/create.go
```diff
@@ -60,10 +78,6 @@ func (r *TreeRepository) createEntity(ctx context.Context, entity *entities.Tree
 	var sensorID *int32
 	if entity.Sensor != nil {
 		sensorID = &entity.Sensor.ID
```
doriengr, line 80: Please do not delete the `UnlinkSensorIDFromTrees` here! Please also write a successful and error test case for this.

internal/storage/postgres/tree/get_test.go
```diff
@@ -0,0 +1,662 @@
...
+		// when
+		treeCluster, err := r.GetTreeClusterByTreeID(ctx, 1)
+
+		// then
+		assert.Error(t, err)
+		assert.Nil(t, treeCluster)
+	})
+}
+
+func assertExpectedEqualToTree(t *testing.T, expectedTree, tree *entities.Tree) {
```
doriengr, line 605: What about checking the `sensorID` and `treeclusterID` here?

internal/storage/postgres/tree/trees.go
```diff
@@ -122,15 +122,26 @@ func (r *TreeRepository) Delete(ctx context.Context, id int32) error {
...
 		}
 	}
-	return r.store.DeleteTree(ctx, id)
+
+	rowID, err := r.store.DeleteTree(ctx, id)
+	if err != nil {
+		return err
+	}
+
+	if rowID != id || rowID == 0 {
```
doriengr, line 140:
```
	if rowID != id || rowID == 0 {
		return storage.ErrTreeNotFound
	}
```

This can be deleted, it's not needed.

internal/storage/postgres/tree/trees.go
```diff
@@ -122,15 +122,26 @@ func (r *TreeRepository) Delete(ctx context.Context, id int32) error {
...
 			return r.store.HandleError(errors.Wrap(err, "failed to unlink image"))
 		}
 
 		if err = r.store.DeleteImage(ctx, img.ID); err != nil {
 			return r.store.HandleError(errors.Wrap(err, "failed to delete image"))
 		}
 	}
-	return r.store.DeleteTree(ctx, id)
+
+	rowID, err := r.store.DeleteTree(ctx, id)
```
doriengr, line 135: The rowId is not needed so use 

`_, err := r.store.DeleteTree(ctx, id)`

internal/storage/postgres/tree/update.go
```diff
@@ -32,35 +32,51 @@ func (r *TreeRepository) UpdateWithImages(ctx context.Context, id int32, tFn ...
...
-
 	var sensorID *int32
 	if t.Sensor != nil {
 		sensorID = &t.Sensor.ID
+	}
 
-		if err := r.store.UnlinkSensorIDFromTrees(ctx, sensorID); err != nil {
-			return err
-		}
+	var treeClusterID *int32
```
doriengr, line 77: Please do not delete `UnlinkSensorIDFromTrees`! Maybe you can also add a test case for this.

internal/storage/storage.go
```diff
@@ -30,6 +30,10 @@ var (
 	ErrConnectionClosed = errors.New("connection is closed")
 	ErrTxClosed         = errors.New("transaction closed")
 	ErrTxCommitRollback = errors.New("transaction cannot commit or rollback")
+
+	ErrInvalidLatitude         = errors.New("latitude must be between -190,180")
```
doriengr, line 34: Are these error messages correct? Why -190?

doriengr (changes requested):

internal/storage/postgres/tree/create.go
Mohammad96Assaf: Hello Dorien, sorry for the mistake. I accidentally removed the function call while copy&past from the old branch.

I tried to test the function with the following test, but somehow the function doesn't work correctly.
Here is my test and the result:

```go
t.Run("should unlink the sensor id in other trees", func(t *testing.T) {
		// given
		suite.ResetDB(t)
		suite.InsertSeed(t, "internal/storage/postgres/seed/test/tree")
		r := NewTreeRepository(suite.Store, mappers)

		sqlSensor, sensorErr := suite.Store.GetSensorByID(context.Background(), 1)
		if sensorErr != nil {
			t.Fatal(sensorErr)
		}
		sensor := mappers.sMapper.FromSql(sqlSensor)

		existTree, existTreeErr := r.GetByID(context.Background(), 1)
		if existTreeErr != nil {
			t.Fatal(sensorErr)
		}
		assert.NotNil(t, existTree.Sensor.ID)
		assert.Equal(t, sensor.ID, existTree.Sensor.ID)

		// when
		got, err := r.Create(context.Background(),
			WithSpecies("Oak"),
			WithTreeNumber("T001"),
			WithPlantingYear(2023),
			WithLatitude(54.801539),
			WithLongitude(9.446741),
			WithDescription("A newly planted oak tree"),
			WithWateringStatus(entities.WateringStatusGood),
			WithSensor(sensor),
		)

		sensorByNewTree, errSensorByTree := r.GetSensorByTreeID(context.Background(), got.ID)

		// then
		assert.NoError(t, err)
		assert.NotNil(t, got)
		assert.NotZero(t, got.ID)
		assert.WithinDuration(t, got.CreatedAt, time.Now(), time.Second)
		assert.WithinDuration(t, got.UpdatedAt, time.Now(), time.Second)
		assert.NoError(t, errSensorByTree)
		assert.NotNil(t, sensorByNewTree)
		assert.Equal(t, sensor.ID, sensorByNewTree.ID)
		assert.Nil(t, existTree.Sensor.ID)
```
The result:

```
2024/11/18 22:14:16 ERROR An Error occurred in database operation error="treecluster not found"
    create_test.go:138:
                Error Trace:    ../green-ecolution-backend/internal/storage/postgres/tree/create_test.go:138
                Error:          Expected nil, but got: 1
                Test:           TestTreeRepository_Create/should_unlink_the_sensor_id_in_other_trees
```
I expect that the ID in the old tree should be removed, but it still exists!

Please check my test above and if necessary, fix the function so that we can test it properly. Let me know if my test might be incorrect. Thank you!

internal/storage/postgres/tree/get_test.go
Mohammad96Assaf: I test it before calling the function because I need to first get the objects from the database.

```go
		assert.NoError(t, err)
		assert.NotNil(t, tree)
		assert.Equal(t, tree.TreeCluster.ID, treeCluster.ID)
		assert.Equal(t, tree.Sensor.ID, sensor.ID)
		assert.NotNil(t, tree.Sensor)
		assertExpectedEqualToTree(t, tree, testTrees[0])
```
		
please check  `TestTreeRepository_GetByID`

internal/storage/storage.go
Mohammad96Assaf: oh! Yes you are right, the latitude muss be between 90 and -90.

internal/storage/postgres/tree/get_test.go
doriengr: I would like to have everything in one function, but it's okay 👍 I can live with it.

internal/storage/postgres/tree/create.go
doriengr: Ignore the test and please create a new issue.

internal/storage/postgres/tree/create.go
```diff
@@ -114,3 +131,25 @@ func (r *TreeRepository) linkImages(ctx context.Context, treeID, imgID int32) er
...
+	}
+	params := sqlc.GetTreeByCoordinatesParams{
+		Latitude:  tree.Latitude,
+		Longitude: tree.Longitude,
+	}
+	treeByCoordinates, err := r.store.GetTreeByCoordinates(ctx, &params)
+
+	if err == nil && treeByCoordinates != nil {
+		return storage.ErrTreeWithSameCoordinates
+	}
```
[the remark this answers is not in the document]
Mohammad96Assaf: What about the import logic we built in release one? We need the function `GetTreeByCoordinates` for the import functionality to check if a tree with the same coordinates exists. Please have a look at `\internal\service\domain\tree\import_tree.go`.

We assumed that the coordinates can be mapped to trees.

internal/storage/storage.go
```diff
@@ -84,8 +88,10 @@ type TreeRepository interface {
 	GetSensorByTreeID(ctx context.Context, id int32) (*entities.Sensor, error)
 	GetTreesByIDs(ctx context.Context, ids []int32) ([]*entities.Tree, error)
 	GetByCoordinates(ctx context.Context, latitude, longitude float64) (*entities.Tree, error)
+	GetTreeClusterByTreeID(ctx context.Context, treeID int32) (*entities.TreeCluster, error)
 
 	UpdateWithImages(ctx context.Context, id int32, fFn ...entities.EntityFunc[entities.Tree]) (*entities.Tree, error)
+	UpdateTreeClusterID(ctx context.Context, treeIDs []int32, treeClusterID *int32) error
```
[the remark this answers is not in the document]
Mohammad96Assaf: Yes, because we need to test it.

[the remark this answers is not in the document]
choffmann: Yes, but not in the service. It looks like it's just a function inside the repo, so it should be private and not in the interface.

internal/storage/postgres/tree/create.go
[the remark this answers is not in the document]
choffmann: OK, that's a point, but it's being moved to the plugin system so the plugin will have to deal with it. However, my point is still that there can be trees or sensors with the same coordinates. Just because the TBZ can't give us an ID for the trees, we shouldn't rely on it being unique. Coordinates are always more accurate depending on how many decimals you use. Another argument is, what if a tree, sensor or tree cluster is archived? Then a new object can be created with the same coordinates.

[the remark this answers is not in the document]
Mohammad96Assaf: Okay, we have to discuss these points in our weekly meeting. For now, I cannot remove the function because I need to change other points that are unrelated to the purpose of this MR (testing the tree repository). We can create another issue for it.

choffmann:
Please remove this file as well 
image

internal/storage/postgres/tree/trees.go
```diff
@@ -122,15 +122,22 @@ func (r *TreeRepository) Delete(ctx context.Context, id int32) error {
 			TreeID:  id,
 			ImageID: img.ID,
 		}
-		if err = r.store.UnlinkTreeImage(ctx, &args); err != nil {
+		_, err = r.store.UnlinkTreeImage(ctx, &args)
+		if err != nil {
```
choffmann, line 126: Just a small note, you can use it here as before in an if condition. You do not need to change this, just a little hint because I actually like this style.

```go
if _, err = r.store.UnlinkTreeImage(ctx, &args); err != nil {
  return r.store.HandleError(errors.Wrap(err, "failed to unlink image"))
}
```

choffmann (approved):

doriengr (approved):

Issue #85, opened 2024-09-29
Implement transactions


doriengr:
Example:
If a treecluster is deleted, the treecluster ID is removed from the trees beforehand and set to null. The treecluster is then deleted. However, as soon as an error occurs in the delete function, the treecluster ID should be added again to the linked trees.

This behaviour is possible with transactions in Go and should be implemented to prevent errors.

### Missing transactions:
- [x] treecluster
- [x] tree
- [x] sensor
- [ ] import 
- [x] vehicles
- [x] watering plan

PR #235, opened 2024-11-18
feat: implement transaction for tree cluster

choffmann:
Ref #85 

my suggestion for the implementation of transactions. In this PR only transactions for treecluster were implemented

- refactor treecluster repo and service
- refactor geo locator
- implement transaction
- change store implementation
- add unit test for store

### DIFF
--- a/internal/service/domain/treecluster/treecluster.go
+++ b/internal/service/domain/treecluster/treecluster.go
@@ -97,33 +97,35 @@ func (s *TreeClusterService) Create(ctx context.Context, tc *domain.TreeClusterC
 	return c, nil
 }
 
-func (s *TreeClusterService) Update(ctx context.Context, id int32, tc *domain.TreeClusterUpdate) (*domain.TreeCluster, error) {
-	if err := s.validator.Struct(tc); err != nil {
-		return nil, service.NewError(service.BadRequest, errors.Wrap(err, "validation error").Error())
-	}
+func (s *TreeClusterService) Update(ctx context.Context, id int32, tcUpdate *domain.TreeClusterUpdate) (*domain.TreeCluster, error) {
+	err := s.treeClusterRepo.Update(ctx, id, func(tc *domain.TreeCluster) (bool, error) {
+		if err := s.validator.Struct(tc); err != nil {
+			return false, service.NewError(service.BadRequest, errors.Wrap(err, "validation error").Error())
+		}
 
-	trees, err := s.getTrees(ctx, tc.TreeIDs)
+		trees, err := s.getTrees(ctx, tcUpdate.TreeIDs)
+		if err != nil {
+			return false, err
+		}
+
+		tc.Trees = trees
+		tc.Name = tcUpdate.Name
+		tc.Address = tcUpdate.Address
+		tc.Description = tcUpdate.Description
+		tc.SoilCondition = tcUpdate.SoilCondition
+
+		if err = s.locator.UpdateCluster(ctx, tc); err != nil {
+			return false, err
+		}
+
+		return true, nil
+	})
+
 	if err != nil {
 		return nil, handleError(err)
 	}
 
-	c, err := s.treeClusterRepo.Update(ctx, id,
-		treecluster.WithTrees(trees),
-		treecluster.WithName(tc.Name),
-		treecluster.WithAddress(tc.Address),
-		treecluster.WithDescription(tc.Description),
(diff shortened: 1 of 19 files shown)


internal/service/domain/treecluster/treecluster.go
```diff
@@ -54,76 +53,69 @@ func (s *TreeClusterService) GetByID(ctx context.Context, id int32) (*domain.Tre
...
 
-	visitedClusters := make(map[int32]bool)
-	for _, tree := range trees {
-		if tree.TreeCluster != nil && tree.TreeCluster.ID != 0 {
-			if _, ok := visitedClusters[tree.TreeCluster.ID]; ok {
-				slog.Debug("Tree already visited", "treeID", tree.ID)
-				continue
-			}
+	c, err := s.treeClusterRepo.Create(ctx, func(tc *domain.TreeCluster) (bool, error) {
+		if err = s.handlePrevTreeLocation(ctx, trees); err != nil {
```
doriengr, line 67: The code looks way cleaner 🔥

internal/service/domain/treecluster/treecluster.go
```diff
@@ -149,8 +141,54 @@ func (s *TreeClusterService) Ready() bool {
 	return s.treeClusterRepo != nil
 }
 
+// handlePrevTreeLocation updates the locations of clusters associated with the provided trees.
```
doriengr, line 144: I quickly googled and I think comments with `/* */` are also possible. Maybe that will make it look a little cleaner.

```
/*
 	handlePrevTreeLocation updates the locations of clusters associated with the provided trees.
	This method iterates over a list of trees and processes the clusters they belong to.
	For each cluster, the cluster's location is updated by recalculating its coordinates and region using the `GeoClusterLocator`.
	Clusters are only updated once, even if multiple trees belong to the same cluster.

	Parameters:
		- ctx: The context for managing request-scoped values, cancellation, and timeouts.
		- trees: A slice of Tree entities to process. Each tree may reference a cluster.

	Returns:
		An error if any cluster update fails, otherwise nil.

	Notes:
		- Clusters with an ID of 0 are ignored.
		- Updates are performed via a callback mechanism in the `treeClusterRepo` to ensure thread safety or transactional consistency.
*/

```

doriengr: We should also check which other functions could use some doc comments.

internal/service/service.go
```diff
@@ -88,7 +88,20 @@ type TreeClusterService interface {
 }
 
 type GeoClusterLocator interface {
-	UpdateCluster(ctx context.Context, clusterID *int32) error
+	// UpdateCluster updates the center coordinates and region of the specified TreeCluster based on its contained trees.
```
doriengr, line 91: Same with comment structure here.

internal/storage/storage.go
```diff
@@ -70,7 +75,20 @@ type VehicleRepository interface {
 }
 
 type TreeClusterRepository interface {
-	BasicCrudRepository[entities.TreeCluster]
+	// GetAll returns all tree clusters
```
doriengr, line 78: I think the comments are a bit out of place here. I think the comments are better placed where the function is also created, so that you can also specify parameters and return values if required. Otherwise this file will get messy.

doriengr (changes requested):

internal/service/domain/treecluster/treecluster.go
choffmann: Yes, this is my start on adding doc comments. As for the `/* */` thing, I would prefer the `//` style, because even the default lib in go uses this style to document the code.

Example in std http lib
image

doriengr: Hmm … than let's do it the ugly way

internal/storage/storage.go
choffmann: Yes, but the problem is that the service uses this interface, if you specify the comment in the implementation, the LSP will not register it because it has no idea of the implementation, only the interface. So it has to be in the interface definitions

doriengr: I do not get it. What is a LSP?

choffmann: LSP stands for [`Language Server Protocol`](https://microsoft.github.io/language-server-protocol/). If I comment the function, it should be visible in the function definition. But because we are using the interface in the service layer, only these comments are visible, not the comments in the functions.

**Comment on interface visible in the service**
image

**Comment on the funtion not visible in the service**
image

PR #239, opened 2024-11-21
Feature/edit database for sensors

### LINKED ISSUE #233: Edit database for sensors
- [x] Remove id (PK) from sensors
- [x] Add device_id (PK) to sensors as text
- [x] Remove sensor_id from sensor_data
- [x] Add foreign key device_id from sensors to sensor_data
- [x] Add latitude, longitude and geometry to sensors
- [x] Add tests for the new attributes
- [x] Edit data (JSON) inside sensor_data (for one box)
```
{
  temperature: 20,
  humidity: 50,
  battery: 34,
  watermarks: [
    {
      resistance: 23,
      centibar: 38,
      depth: 30
    },
    {
      resistance: 23,
      centibar: 38,
      depth: 60
    },
    {
      resistance: 23,
      centibar: 38,
      depth: 90
    }
  ]
}
```

- [x] Also edit Watermark struct
```go
type Watermark struct {
    Centibar   int
    Resistance int
    Depth int
}
```

- [x] Also edit Payload struct

```go
type MqttDecodedPayload struct {
    Device_id string
    Battery     float64
    Humidity    int
    Temperature string
    Watermarks  []Watermark
}
```

Mohammad96Assaf:
close #233


internal/storage/postgres/migrations/20240825201224_release_1.sql
choffmann, line 1: Do not edit this migration. Instead, create a new migration to edit the database definition.

internal/entities/sensor.go
```diff
@@ -11,26 +11,33 @@ const (
...
 	UpdatedAt time.Time
 	Status    SensorStatus
 	Data      []*SensorData
+	Latitude  float64
+	Longitude float64
 }
 
 type SensorData struct {
 	ID        int32
+	SensorID  *string
```
choffmann, line 25: We do not need the ID here because `Sensor` has an array of SensorData

internal/storage/postgres/sensor/get.go
```diff
@@ -69,3 +71,22 @@ func (r *SensorRepository) GetSensorDataByID(ctx context.Context, id int32) ([]*
 
 	return domainData, nil
 }
+
+func (r *SensorRepository) GetCenterPoint(ctx context.Context, ids []int32) (lat, long float64, err error) {
```
choffmann, line 75: Why do we need this function?

choffmann (changes requested):

internal/storage/postgres/migrations/20240825201224_release_1.sql
Mohammad96Assaf: It was a good point to practice handling migrations in the project 👍

internal/storage/postgres/sensor/get.go
Mohammad96Assaf: I thought we needed it just like in tree ?

internal/entities/sensor.go
Mohammad96Assaf: What about the function `InsertSensorData()` ?

Mohammad96Assaf: I think there was an error there, you used the ID of the sensor data as the ID for the sensor:
```go
	params := &sqlc.InsertSensorDataParams{
			SensorID: d.ID,
			Data:     raw,
		}
```

Mohammad96Assaf: Have been discussed.

internal/storage/postgres/sensor/get.go
Mohammad96Assaf: Have been discussed.

internal/storage/postgres/queries/sensors.sql
```diff
@@ -12,22 +12,34 @@ SELECT * FROM sensor_data WHERE sensor_id = $1;
...
+  sensor_id, data
 ) VALUES (
   $1, $2
 ) RETURNING id;
 
+-- name: UpdateSensorGeometry :exec
+UPDATE sensors SET
+    geometry = ST_GeomFromText($2, 4326)
+WHERE id = $1;
+
```
choffmann, line 43: I dont think we need this when we have `SetSensorLocation`

internal/storage/postgres/seed/test/demo_data/99_demo_data.sql
```diff
@@ -88,15 +87,15 @@ VALUES
...
-  (3, '{"temperature": 22.0, "humidity": 0.7, "moisture": 0.25}');
+  ('sensor-1', '{"temperature": 20.0, "humidity": 0.5, "moisture": 0.75}'),
+  ('sensor-1', '{"temperature": 21.0, "humidity": 0.6, "moisture": 0.5}'),
+  ('sensor-1', '{"temperature": 22.0, "humidity": 0.7, "moisture": 0.25}'),
+  ('sensor-2', '{"temperature": 20.0, "humidity": 0.5, "moisture": 0.75}'),
+  ('sensor-2', '{"temperature": 21.0, "humidity": 0.6, "moisture": 0.5}'),
+  ('sensor-2', '{"temperature": 22.0, "humidity": 0.7, "moisture": 0.25}'),
+  ('sensor-3', '{"temperature": 20.0, "humidity": 0.5, "moisture": 0.75}'),
+  ('sensor-3', '{"temperature": 21.0, "humidity": 0.6, "moisture": 0.5}'),
+  ('sensor-3', '{"temperature": 22.0, "humidity": 0.7, "moisture": 0.25}');
```
choffmann, line 98: This is the old structure. Or is the test using this data? If so, we should update the test

internal/storage/postgres/seed/test/demo_data/99_demo_data.sql
```diff
@@ -19,18 +19,17 @@ VALUES
...
 
-INSERT INTO sensors (id, status) VALUES (1, 'online');
-INSERT INTO sensors (id, status) VALUES (2, 'offline');
-INSERT INTO sensors (id, status) VALUES (3, 'unknown');
-INSERT INTO sensors (id, status) VALUES (4, 'online');
-ALTER SEQUENCE sensors_id_seq RESTART WITH 5;
+INSERT INTO sensors (id, status) VALUES ('sensor-1', 'online');
+INSERT INTO sensors (id, status) VALUES ('sensor-2', 'offline');
+INSERT INTO sensors (id, status) VALUES ('sensor-3', 'unknown');
+INSERT INTO sensors (id, status) VALUES ('sensor-4', 'online');
```
choffmann, line 25: The demo data in the test should also have the new structure from the sensors

choffmann (changes requested):

internal/storage/postgres/seed/test/demo_data/99_demo_data.sql
Mohammad96Assaf: Please take a look at the latest update of the file: 
https://github.com/green-ecolution/green-ecolution-backend/blob/feature/edit-database-for-sensors/internal/storage/postgres/seed/99_demo_data.sql

internal/storage/postgres/seed/test/demo_data/99_demo_data.sql
Mohammad96Assaf: Please take a look at the latest update of the file: 
https://github.com/green-ecolution/green-ecolution-backend/blob/feature/edit-database-for-sensors/internal/storage/postgres/seed/99_demo_data.sql

internal/storage/postgres/seed/test/demo_data/99_demo_data.sql
choffmann: Yes, and now? This is a different file for a different use case. This is for testing. But I don't see this file being used at all in the project, so please delete it.

Mohammad96Assaf: Yes, you are right 👍

internal/server/http/handler/v1/tree/handler_test.go
```diff
@@ -102,32 +102,33 @@ func TestGetTreeBySensorID(t *testing.T) {
...
+	//
+	//	// when
+	//	req, _ := http.NewRequestWithContext(context.Background(), http.MethodGet, "/v1/tree/sensor/"999, nil)
+	//	resp, err := app.Test(req, -1)
+	//	defer resp.Body.Close()
+	//
+	//	// then
+	//	assert.Nil(t, err)
+	//	assert.Equal(t, http.StatusBadRequest, resp.StatusCode)
+	//})
```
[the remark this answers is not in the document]
Mohammad96Assaf: I implemented a function to test the if the sensorID is invalid or not. you can review it 👍

internal/server/http/handler/v1/tree/handler.go
```diff
@@ -90,10 +107,36 @@ func GetTreeByID(svc service.TreeService) fiber.Handler {
 // @Security		Keycloak
 func GetTreeBySensorID(svc service.TreeService) fiber.Handler {
 	return func(c *fiber.Ctx) error {
-		return getTreeByField(c, "sensor_id", svc.GetBySensorID)
+		ctx := c.Context()
+		id := c.Params("sensor_id")
```
choffmann, line 111: Please clone the value of `c.Params`, as I am currently having major problems with this exact error
```suggestion
		id := strings.Clone(c.Params("sensor_id"))
```

internal/server/http/handler/v1/tree/handler.go
```diff
@@ -90,10 +107,36 @@ func GetTreeByID(svc service.TreeService) fiber.Handler {
...
+	invalidLength := len(sensorID) < 3 || len(sensorID) > 36
+	hasWhiteSPaces := strings.ContainsAny(sensorID, " \t\n")
+	htmlPattern := regexp.MustCompile(`(?i)<[a-z]+[^>]*>.*</[a-z]+>|<.*>`)
+	isLikeHTML := htmlPattern.MatchString(sensorID)
+	if containsSpChars || invalidLength || hasWhiteSPaces || isLikeHTML {
+		return false
+	}
+
+	return true
+}
```
choffmann, line 138: I don't know if we need this level of detail about a valid sensor ID. I think a simple regex expression for our sensor id type is sufficient. We should also make this a global function inside the sensor entity if you really want to check if the sensor id is valid.

Also, why is this function only used here to validate the sensor ID, not in the sensor handler it self?

choffmann (changes requested):
This PR is getting bigger and bigger. This PR should just fix the database and the entities, not necessarily the logic, but you're adding more and more logic, like sensor ID validation. This PR should go through as quickly as possible so that we can get on with analysing the sensor data.

internal/server/http/handler/v1/tree/handler.go
Mohammad96Assaf: What do think the validation should look like (simple regex)?
I believe that the validation is very important, especially for a string-id.

Mohammad96Assaf:
@choffmann  The PR has grown bigger because of the new functions and tests from the develop branch, which use the **old** **sensorID**.
However, I have removed the validation function, and I think this PR is now ready to be merged.

choffmann (approved):
:+1:

PR #243, opened 2024-11-22
Feature/update vehicle entities

### LINKED ISSUE #236: Update vehicle entities and database table with new fields
- [x] height
- [x] width
- [x] length
- [x] needed driving licence
- [x] name/model

doriengr:
Close #236


internal/storage/postgres/vehicle/create.go
```diff
@@ -10,11 +10,16 @@ import (
...
-		Description:   "",
-		WaterCapacity: 0,
-		Type:          entities.VehicleTypeUnknown,
-		Status:        entities.VehicleStatusUnknown,
+		NumberPlate:    "",
+		Description:    "",
+		WaterCapacity:  0,
+		Type:           entities.VehicleTypeUnknown,
+		Status:         entities.VehicleStatusUnknown,
+		Model:          "",
```
choffmann, line 18: In your migration you have set the default model type to `Unknown`. Here you set the default to an empty string. If you now create a vehicle with this default type and do not specify a model, the model in the database will be an empty string. Either set the default model here to `Unknown` as in the migration, or remove the default value in the table definition.

choffmann (changes requested):

choffmann (approved):
Approved when the pipelines are fixed :+1:

PR #244, opened 2024-11-23
Feature/storage layer for watering plan

doriengr:
refs #230 

I haven't created any great logic yet. For now, it's just about creating basic CRUD functions that can be expanded. Otherwise, the MR will immediately become large and confusing.


internal/storage/postgres/watering_plan/update.go
```diff
@@ -0,0 +1,51 @@
...
+	if err := w.validateWateringPlan(entity); err != nil {
+		return nil, err
+	}
+
+	if err := w.updateEntity(ctx, entity); err != nil {
+		return nil, err
+	}
+
+	return w.GetByID(ctx, entity.ID)
+}
```
choffmann, line 31: Maybe we can already use the transaction here (not tested the suggestion)

```suggestion
func (w *WateringPlanRepository) Update(ctx context.Context, id int32, updateFn func(*entities.WateringPlan) (bool, error)) error {
	return w.store.WithTx(ctx, func(s *store.Store) error {
		oldStore := r.store
		defer func() {
			r.store = oldStore
		}()
		r.store = s

		entity, err := w.GetByID(ctx, id)
		if err != nil {
			return err
		}

		if updateFn == nil {
			return errors.New("updateFn is nil")
		}

		updated, err := updateFn(entity)
		if err != nil {
			return err
		}

		if !updated {
			return nil
		}

		return w.updateEntity(ctx, tc)
	})

}
```

internal/storage/postgres/watering_plan/create.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	id, err := w.createEntity(ctx, entity)
+	if err != nil {
+		return nil, w.store.HandleError(err)
+	}
+
+	entity.ID = *id
+	return w.GetByID(ctx, *id)
+}
```
choffmann, line 43: Maybe we can already use the transaction here (not tested the suggestion)
```suggestion
func (w *WateringPlanRepository) Create(ctx context.Context, createFn func(*entities.WateringPlan) (bool, error)) (*entities.WateringPlan, error) {
	if createFn == nil {
		return nil, errors.New("createFn is nil")
	}
	
	var createdW *entities.WateringPlan
	err := w.store.WithTx(ctx, func(s *store.Store) err {
		oldStore := r.store
		defer func() {
			r.store = oldStore
		}()
		r.store = s
		
		entity := defaultWateringPlan()
		created, err := createFn(entity)
		if err != nil {
			return err
		}

		if !created {
			return nil
		}

		if err := w.validateWateringPlan(entity); err != nil {
			return nil, err
		}

		id, err := w.createEntity(ctx, entity)
		if err != nil {
			return nil, w.store.HandleError(err)
		}
		
		createdW, err = w.GetByID(ctx, *id)
		if err != nil {
			return err
		}
		
		return nil
	}
	return createdW, nil
}
```

internal/storage/postgres/migrations/20241116142300_implement_watering_plans.sql
```diff
@@ -0,0 +1,62 @@
...
+CREATE OR REPLACE FUNCTION update_updated_at_column()
+  RETURNS TRIGGER
+  AS $$
+BEGIN
+    NEW.updated_at = NOW();
+    RETURN NEW;
+END;
+$$
+language 'plpgsql';
+-- +goose StatementEnd
```
choffmann, line 48: You dont need this here, the function already exists

choffmann (changes requested):

choffmann (approved):
:+1:

PR #257, opened 2024-12-04
Teature/middleware test

ShadowDataNerd:
Tests for the NewJWTMiddleware are still missing and need to be added.


doriengr:
The tests are failing and the linter is also failing. Is it correct that this is ready for review?

internal/server/http/middleware/health_check_test.go
```diff
@@ -0,0 +1,133 @@
...
+	handler := healthcheck.New(healthcheck.Config{
+		LivenessProbe: func(_ *fiber.Ctx) bool {
+			return true
+		},
+		LivenessEndpoint: "/health",
+		ReadinessProbe: func(_ *fiber.Ctx) bool {
+			return true
+		},
+		ReadinessEndpoint: "/ready",
+	})
```
choffmann, line 26: Do not redefine the handler here. Use the actual defined handler.

internal/server/http/middleware/health_check_test.go
```diff
@@ -0,0 +1,133 @@
...
+	handler := healthcheck.New(healthcheck.Config{
+		LivenessProbe: func(_ *fiber.Ctx) bool {
+			return true
+		},
+		LivenessEndpoint: "/health",
+		ReadinessProbe: func(_ *fiber.Ctx) bool {
+			return false
+		},
+		ReadinessEndpoint: "/ready",
+	})
```
choffmann, line 66: Same here, please do not create a new handler

internal/server/http/middleware/requestid_test.go
choffmann, line 1: This is the same as the logger, just using fibre middleware with no logic of its own. Although test cases are used here, many of them have nothing to do with this. I would perhaps only use the test case `should add X-Request-ID header to each request` and `should generate unique IDs for different requests`.

choffmann (changes requested):
Please fix the linter and my annotations, I will work on the JWT tests

internal/server/http/middleware/health_check_test.go
choffmann: I just have implemented it

internal/server/http/middleware/health_check_test.go
```diff
@@ -0,0 +1,126 @@
...
+		assert.Equal(t, http.StatusOK, resp.StatusCode)
+	})
+
+	t.Run("should return 404 for undefined endpoint", func(t *testing.T) {
+		req := httptest.NewRequest(http.MethodGet, "/undefined", nil)
+		resp, err := app.Test(req, -1)
+
+		assert.NoError(t, err)
+		assert.Equal(t, http.StatusNotFound, resp.StatusCode)
+	})
```
choffmann, line 44: We need tests here if the application is not ready. Please mock `AllServicesReady()` and return `false` and test it. The readiness probe should not return 200.

internal/server/http/middleware/health_check_test.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+func TestHealthCheck_EdgeCases(t *testing.T) {
```
choffmann, line 47: I think we don't need that either, these are all edge cases that we don't need to test on the side of this middleware.

choffmann (changes requested):
Just some more feedback on `health_check`

choffmann:
@ShadowDataNerd I have implemented the tests for the jwt middleware. Please fix the linter and my comments :+1:

choffmann (approved):

PR #259, opened 2024-12-06
Feature/mapping sensor to tree

### LINKED ISSUE #220: Implement mapping logic: sensor data to tree

Mohammad96Assaf:
close  #220


internal/service/domain/sensor/sensor.go
```diff
@@ -100,6 +104,30 @@ func (s *SensorService) Delete(ctx context.Context, id string) error {
...
 }
 
+func (s *SensorService) MapSensorToTree(ctx context.Context, sen *entities.Sensor) error {
+	if sen == nil {
+		return errors.New("sensor cannot be nil")
+	}
+
+	if err := s.validator.Struct(sen); err != nil {
+		return service.NewError(service.BadRequest, errors.Wrap(err, "validation error").Error())
+	}
```
choffmann, line 114: OK, but what do you have to validate if the entity has no annotations?

choffmann: And in this case, a sensor should already be defined, existing and validated, or not?

internal/service/domain/sensor/sensor_test.go
```diff
@@ -431,6 +433,87 @@ func TestSensorService_Delete(t *testing.T) {
...
+
+		// when
+		err := svc.MapSensorToTree(context.Background(), nil)
+
+		// then
+		assert.Error(t, err)
+		assert.EqualError(t, err, "sensor cannot be nil")
+	})
+
+	t.Run("should return error if nearest tree is not found", func(t *testing.T) {
```
doriengr, line 475: I may be wrong, but I am wondering: if a sensor cannot be assigned to a tree because there is no tree nearby, why is an error returned? For example, if we call this function when importing trees, the import should not stop but continue because the sensor that could not be mapped does not have a linked tree.

But as I said, I do not know if I got the assignment wrong.

internal/storage/postgres/queries/trees.sql
```diff
@@ -83,3 +83,9 @@ UPDATE trees SET sensor_id = NULL WHERE sensor_id = $1;
 
 -- name: CalculateGroupedCentroids :one
 SELECT ST_AsText(ST_Centroid(ST_Collect(geometry)))::text AS centroid FROM trees WHERE id = ANY($1::int[]);
+
+-- name: FindNearestTree :one
+SELECT * FROM trees
+WHERE ST_Distance(geometry::geography, ST_SetSRID(ST_MakePoint($1, $2), 4326)::geography) <= 3
+ORDER BY ST_Distance(geometry::geography, ST_SetSRID(ST_MakePoint($1, $2), 4326)::geography) ASC
+    LIMIT 1;
```
doriengr, line 91: Missing empty line at the bottom 👍

internal/storage/postgres/tree/get_test.go
```diff
@@ -664,7 +664,64 @@ func TestTreeRepository_GetTreeClusterByTreeID(t *testing.T) {
 		assert.Nil(t, treeCluster)
 	})
 }
+func TestTreeRepository_FindNearestTree(t *testing.T) {
```
doriengr, line 667: There is a missing line at the top 👍

internal/storage/postgres/tree/get_test.go
```diff
@@ -664,7 +664,64 @@ func TestTreeRepository_GetTreeClusterByTreeID(t *testing.T) {
...
+		// when
+		nearestTree, errFind := r.FindNearestTree(context.Background(), sensorLatitude, sensorLongitude)
+
+		// then
+		assert.NoError(t, errFind, "Expected no error while finding the nearest tree")
+		assert.NotNil(t, nearestTree, "Expected to find a nearest tree")
+		assertExpectedEqualToTree(t, tree, nearestTree)
+	})
+
+	t.Run("Should return error when no tree found within the required distance", func(t *testing.T) {
```
doriengr, line 691: "should"

internal/service/domain/sensor/sensor_test.go
Mohammad96Assaf: I think we could adjust the import function to continue its work if no tree is found for a given sensor.

Also, I'm not sure if we are calling this function during the import process, as the parameter here is the sensor itself. In the import process, we are searching for the sensor, and the parameter should be the imported tree instead.

With my implementation, my main goal was to process the sensor data from TTN. I think we should discuss this point.

doriengr: You are right. Import is not correct, but creating a sensor from the TTN. I am approving it 👍

doriengr (approved):

choffmann (approved):

PR #262, opened 2024-12-07
Feature/sensor status updater

### LINKED ISSUE #256: Implement Sensor Status Updater Based on Last Update Time
We need to create a function that runs continuously to check when each sensor was last updated. If a sensor has not been updated for more than 3 days, then we change its status automatically to `offline`.

Mohammad96Assaf:
close #256


internal/service/domain/sensor/sensor.go
```diff
@@ -21,11 +23,15 @@ func NewSensorService(
...
 	flowerbedRepo storage.FlowerbedRepository,
 ) service.SensorService {
-	return &SensorService{
+	s := &SensorService{
 		sensorRepo:    sensorRepo,
 		treeRepo:      treeRepo,
 		flowerbedRepo: flowerbedRepo,
 	}
+
+	go s.RunStatusUpdater(context.Background(), 1*time.Hour) // Run every 1 hour
```
choffmann, line 32: I wouldn't run it here because it runs every time you create the service. I would put it in `server.go`, there is also the plugin cleanup service.

internal/service/domain/sensor/sensor.go
```diff
@@ -100,6 +106,46 @@ func (s *SensorService) Delete(ctx context.Context, id string) error {
...
 }
 
+func (s *SensorService) RunStatusUpdater(ctx context.Context, interval time.Duration) {
+	ticker := time.NewTicker(interval)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ticker.C:
+			slog.Info("EVERY two min")
```
choffmann, line 116: Please change the text to something meaningful

internal/service/domain/sensor/sensor.go
```diff
@@ -100,6 +106,46 @@ func (s *SensorService) Delete(ctx context.Context, id string) error {
...
+			err := s.updateStaleSensorStatuses(ctx)
+			if err != nil {
+				slog.Error("Error updating sensor statuses", "error", err.Error())
```
choffmann, line 119: This message sounds a bit odd, how about this

```suggestion
				slog.Error("Failure to update sensor status", "error", err.Error())
```

internal/service/domain/sensor/sensor_test.go
```diff
@@ -366,6 +367,88 @@ func TestSensorService_Delete(t *testing.T) {
...
+		sensorRepo.EXPECT().GetAll(mock.Anything).Return([]*entities.Sensor{staleSensor, recentSensor}, nil)
+		sensorRepo.EXPECT().Update(mock.Anything, staleSensor.ID, mock.Anything).Return(staleSensor, nil)
+
+		go func() {
+			svc.RunStatusUpdater(ctx, 10*time.Millisecond)
+		}()
+
+		time.Sleep(100 * time.Millisecond)
+
+		// No assertions needed since mock expectations handle it
```
choffmann, line 394: Why, I don't think so. Mock expectations will just return what you define here, which is true, but you need to validate that `svc.RunStatusUpdater` is doing the write thing.

internal/service/domain/sensor/sensor.go
choffmann: You can also use the context from the application, rather than from the background.

choffmann (changes requested):

internal/service/domain/sensor/sensor.go
Mohammad96Assaf: Oh yes, it was just for testing.

internal/service/domain/sensor/sensor_test.go
Mohammad96Assaf: How can we do that if runStatusUpdater is only responsible for running the ticker in a loop?

Mohammad96Assaf: Do you have a better Idea?

internal/service/domain/sensor/sensor.go
```diff
@@ -100,6 +106,46 @@ func (s *SensorService) Delete(ctx context.Context, id string) error {
 	return nil
 }
 
+func (s *SensorService) RunStatusUpdater(ctx context.Context, interval time.Duration) {
```
choffmann, line 118: I think I would move this whole function into a separate component that is called and created within the service. That way you encapsulate the function a bit, which is a bit nicer

internal/service/domain/sensor/sensor_test.go
choffmann: Where do you use `sensorRepo.AssertCalled` and `sensorRepo.AssertNotCalled`? I do not see it. Then this would be your validation and okay

choffmann (changes requested):

internal/service/domain/sensor/sensor_test.go
Mohammad96Assaf: How are they not visible to you? Check this link: [internal/service/domain/sensor/sensor_test.go](https://github.com/green-ecolution/green-ecolution-backend/pull/262/files#diff-e7f09c8b46076b203bf75664c16e541c6f6a412fa390fbe4d1a3fccf30bde18e)

internal/service/domain/sensor/statusUpdater.go
```diff
@@ -0,0 +1,60 @@
...
+	"time"
+
+	"github.com/green-ecolution/green-ecolution-backend/internal/entities"
+	"github.com/green-ecolution/green-ecolution-backend/internal/storage"
+	"github.com/green-ecolution/green-ecolution-backend/internal/storage/postgres/sensor"
+)
+
+type StatusUpdater struct {
+	sensorRepo storage.SensorRepository
+}
```
choffmann, line 15: It would be nice if we can make it more generic, but it's okay for now...

internal/service/domain/sensor/statusUpdater_test.go
```diff
@@ -0,0 +1,104 @@
...
+
+		go func() {
+			svc.RunStatusUpdater(ctx, 10*time.Millisecond)
+		}()
+
+		time.Sleep(100 * time.Millisecond)
+
+		sensorRepo.AssertCalled(t, "GetAll", mock.Anything)
+		sensorRepo.AssertCalled(t, "Update", mock.Anything, staleSensor.ID, mock.Anything)
+		sensorRepo.AssertExpectations(t) // Verifies all expectations are met
```
choffmann, line 40: Now there is it, but before there was no `AssertCalled` function called. This is fine :+1:

choffmann: Maybe I was looking at an old version of the file

choffmann (approved):

PR #275, opened 2024-12-09
Feature/test error responses service layer

### LINKED ISSUE #227: Test for correct error responses in service layer

ShadowDataNerd:
Close #227

### DIFF
--- a/internal/service/domain/auth/user_test.go
+++ b/internal/service/domain/auth/user_test.go
@@ -70,8 +70,10 @@ func TestRegisterUser(t *testing.T) {
 		resp, err := svc.Register(context.Background(), &entities.RegisterUser{})
 
 		// then
-		assert.Error(t, err, "500: failed to register user")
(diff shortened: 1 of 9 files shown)


doriengr:
Use sometimes use `assert.Contains`, `assert.Error`, `assert.EqualError` and `assert.Equal`

`assert.Contains` makes sense if you only want to check if a word is inside the error, for example if you want to check the validation error.  Otherwise we should use `assert.Equal`.

internal/service/domain/auth/user_test.go
```diff
@@ -254,6 +261,7 @@ func TestLogoutRequest(t *testing.T) {
 
 		// when
 		assert.NoError(t, err)
+		assert.Equal(t, nil, err)
```
doriengr, line 264: This is the same as line 263. You do not need it 👍

doriengr (approved):

PR #280, opened 2024-12-17
Refactor/rename treenumber to number

doriengr:
I really did not like it that in the storage layer it was `TreeNumber` and in the server layer `Number` and then again in the frontend its `tree.treeNumber`. I unified everything to `number`. After this is merged we also need to update the frontend.


Mohammad96Assaf (approved):
👍

PR #293, opened 2024-12-23
feature/get all users from keycloack

### LINKED ISSUE #267: Get users from keycloack
Implement getall method to get all users from keycloack database.

Mohammad96Assaf:
close #267

### DIFF
--- a/internal/server/http/handler/v1/user/handler.go
+++ b/internal/server/http/handler/v1/user/handler.go
@@ -187,9 +187,30 @@ func parseURL(rawURL string) (*url.URL, error) {
 // @Param			limit	query		string	false	"Limit"
 // @Router			/v1/user [get]
 // @Security		Keycloak
-func GetAllUsers(_ service.AuthService) fiber.Handler {
+func GetAllUsers(svc service.AuthService) fiber.Handler {
 	return func(c *fiber.Ctx) error {
-		return c.SendStatus(fiber.StatusNotImplemented)
+		ctx := c.Context()
+
+		users, err := svc.GetAllUsers(ctx)
+		if err != nil {
+			return errorhandler.HandleError(service.NewError(service.InternalError, errors.Wrap(err, "failed to get all users").Error()))
+		}
+		response := make([]entities.UserResponse, len(users))
+
+		for i, user := range users {
+			response[i] = entities.UserResponse{
+				ID:            user.ID.String(),
+				CreatedAt:     user.CreatedAt,
+				Email:         user.Email,
+				FirstName:     user.FirstName,
+				LastName:      user.LastName,
+				Username:      user.Username,
+				EmployeeID:    user.EmployeeID,
+				PhoneNumber:   user.PhoneNumber,
+				EmailVerified: user.EmailVerified,
+			}
+		}
+		return c.Status(fiber.StatusOK).JSON(response)
 	}
 }
 
--- a/internal/server/http/handler/v1/user/handler.go
+++ b/internal/server/http/handler/v1/user/handler.go
@@ -2,6 +2,7 @@ package user
 
 import (
 	"net/url"
+	"strconv"
 
 	"github.com/gofiber/fiber/v2"
 	"github.com/golang-jwt/jwt/v5"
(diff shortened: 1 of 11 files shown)


internal/server/http/handler/v1/user/handler.go
```diff
@@ -187,9 +188,41 @@ func parseURL(rawURL string) (*url.URL, error) {
 // @Param			limit	query		string	false	"Limit"
 // @Router			/v1/user [get]
 // @Security		Keycloak
-func GetAllUsers(_ service.AuthService) fiber.Handler {
+func GetAllUsers(svc service.AuthService) fiber.Handler {
 	return func(c *fiber.Ctx) error {
-		return c.SendStatus(fiber.StatusNotImplemented)
+		page := c.Query("page")
```
doriengr, line 193: Are you currently using page and limit somewhere? We need those parameters if we want to use pagination, what we are not doing (yet).

internal/service/domain/auth/user.go
```diff
@@ -66,3 +66,12 @@ func (s *AuthService) LogoutRequest(ctx context.Context, logoutRequest *domain.L
 
 	return nil
 }
+
+func (s *AuthService) GetAllUsers(ctx context.Context) ([]*domain.User, error) {
```
doriengr, line 70: In the other services we call this method just »GetAll«. 
To make it uniform everywhere, I would suggest that we name it GetAll here as well? Would that be okay?

internal/server/http/handler/v1/user/handler.go
```diff
@@ -187,9 +188,41 @@ func parseURL(rawURL string) (*url.URL, error) {
...
+		ctx := c.Context()
+
+		users, err := svc.GetAllUsers(ctx)
+		if err != nil {
+			return errorhandler.HandleError(service.NewError(service.InternalError, errors.Wrap(err, "failed to get all users").Error()))
+		}
+		response := make([]entities.UserResponse, len(users))
+
+		for i, user := range users {
+			response[i] = entities.UserResponse{
```
doriengr, line 212: Maybe we should use the http mapper to map the data?

internal/service/domain/auth/user_test.go
```diff
@@ -291,3 +291,61 @@ func TestLogoutRequest(t *testing.T) {
 		assert.Error(t, err)
 	})
 }
+
+func TestGetAllUsers(t *testing.T) {
+	t.Run("should return all users successfully", func(t *testing.T) {
+		// given
```
doriengr, line 297: Maybe add a test case »should return empty slice when user are found«? Same goes for the handler.

internal/storage/auth/keycloak/user.go
```diff
@@ -83,21 +83,50 @@ func (r *UserRepository) RemoveSession(ctx context.Context, refreshToken string)
...
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get users from Keycloak")
+	}
+
+	var allUsers []*entities.User
+
+	for _, kcUser := range users {
+		user, err := keyCloakUserToUser(kcUser)
+		if err != nil {
+			continue
```
doriengr, line 103: Why should it be continued if an error is happening?

internal/storage/auth/keycloak/user_test.go
```diff
@@ -297,3 +297,68 @@ func TestKeyCloakUserRepo_UserToKeyCloakUser(t *testing.T) {
...
+		suite.EnsureUserExists(t, user2)
+
+		// when
+		users, err := userRepo.GetAll(context.Background())
+
+		// then
+		assert.NoError(t, err)
+		assert.NotNil(t, users)
+		assert.GreaterOrEqual(t, len(users), 2)
+		assert.True(t, containsUser(users, *user1), "user1 should be in the list")
```
doriengr, line 333: I am not sure, but can you maybe use `assert.Contain`?

doriengr (changes requested):

internal/server/http/handler/v1/user/handler.go
Mohammad96Assaf: Actually, I wanted to ask the group if we need it in the handler. But I will delete it for now since we don't use it 👍

internal/service/domain/auth/user_test.go
Mohammad96Assaf: I think, when **no** users are found.

internal/storage/auth/keycloak/user.go
Mohammad96Assaf: Continuing would be similar to skipping users that are nil, we don't need to add them to the response list.
It's the same as saying: `if user != nil { allUsers = append(allUsers, user) }`

But what do you suggest, should we break the loop if the conversion from a Keycloak user to user fails and return an error ?

internal/storage/auth/keycloak/user_test.go
Mohammad96Assaf: That's the reason why I implemented the containsUser function. The assert.Contains method checks if the pointer (memory reference) is included or not, which doesn't work in my case

internal/storage/auth/keycloak/user.go
doriengr: I would say so, yes ...

doriengr:
Since I believe that the frontend credentials must be used, the error still needs to be fixed here. I hope that is correct, that the frontend credentials must always be used in Swagger, right? @choffmann 
image

doriengr (approved):

PR #294, opened 2024-12-23
Feature/add sensor data to response

### LINKED ISSUE #192: Add sensor data to sensor response
image

Currently the sensor data is not returned on the get functions. 
Also on create etc the data is not returned

doriengr:
Close #192

### DIFF
--- a/internal/storage/postgres/sensor/entities/mqtt_payload.go
+++ b/internal/storage/postgres/sensor/entities/mqtt_payload.go
@@ -11,5 +11,7 @@ type MqttPayload struct {
 	Battery     float64     `json:"battery"`
 	Humidity    float64     `json:"humidity"`
 	Temperature float64     `json:"temperature"`
+	Latitude    float64     `json:"latitude"`
+	Longitude   float64     `json:"longitude"`
 	Watermarks  []Watermark `json:"watermarks"`
 }
--- a/internal/storage/postgres/sensor/entities/mqtt_payload.go
+++ b/internal/storage/postgres/sensor/entities/mqtt_payload.go
@@ -7,7 +7,7 @@ type Watermark struct {
 }
 
 type MqttPayload struct {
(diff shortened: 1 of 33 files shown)


Mohammad96Assaf (changes requested):
The changes inside **handleMqttMessage** are causing an issue, and the data from TTN cannot be converted correctly.
Please fix the problem and verify if the data is still read from TTN and stored in the database correctly.

doriengr:
> The changes inside **handleMqttMessage** are causing an issue, and the data from TTN cannot be converted correctly. Please fix the problem and verify if the data is still read from TTN and stored in the database correctly.

Csn you show me how to test it?

doriengr:
> The changes inside **handleMqttMessage** are causing an issue, and the data from TTN cannot be converted correctly. Please fix the problem and verify if the data is still read from TTN and stored in the database correctly.

Is fixed 👍

internal/storage/postgres/sensor/entities/mqtt_payload.go
```diff
@@ -7,9 +7,11 @@ type Watermark struct {
 }
 
 type MqttPayload struct {
-	DeviceID    string      `json:"device_id"`
+	Device      string      `json:"device"`
 	Battery     float64     `json:"battery"`
 	Humidity    float64     `json:"humidity"`
 	Temperature float64     `json:"temperature"`
+	Latitude    float64     `json:"latitude"`
+	Longitude   float64     `json:"longitude"`
```
Mohammad96Assaf, line 15: Why are latitude and longitude being added here? These values are already saved once directly as sensor attributes, so there’s no need to include them again in the sensor data, every time the data is stored.

We’ve previously discussed the structure of sensor data in #233.

Mohammad96Assaf (changes requested):

Mohammad96Assaf (approved):

PR #295, opened 2024-12-23
feat: implement event system

choffmann:
Implement an event system. For now, when a tree has been created, updated and deleted, it will send an event which will be subscribed to by the tree cluster to update its locations based on the changed tree.


internal/entities/events.go
```diff
@@ -0,0 +1,74 @@
...
+
+type Event interface {
+	Type() EventType
+}
+
+const (
+	EventTypeUpdateTree        EventType = "update tree"
+	EventTypeCreateTree        EventType = "create tree"
+	EventTypeDeleteTree        EventType = "delete tree"
+	EventTypeUpdateTreeCluster EventType = "update tree cluster"
```
doriengr, line 13: Do we only have to listen on Update for a tree cluster?

internal/service/domain/treecluster/treecluster.go
```diff
@@ -73,10 +107,15 @@ func (s *TreeClusterService) Create(ctx context.Context, createTc *domain.TreeCl
 		tc.Description = createTc.Description
 		tc.Trees = trees
 
-		if err = s.locator.UpdateCluster(ctx, tc); err != nil {
-			return false, handleError(err)
+		lat, long, region, err := s.getUpdatedLatLong(ctx, tc)
```
doriengr, line 110: Maybe put this in a function? From line 151 onwards is the same code

internal/service/service.go
```diff
@@ -81,23 +81,7 @@ type RegionService interface {
 type TreeClusterService interface {
 	Service
 	CrudService[domain.TreeCluster, domain.TreeClusterCreate, domain.TreeClusterUpdate]
-}
-
-type GeoClusterLocator interface {
```
doriengr, line 86: 🔥

doriengr (approved):
I am in love 😍

internal/entities/events.go
choffmann: No, but I would implement each event if it is really needed

PR #297, opened 2024-12-24
Feature/get users by role

### LINKED ISSUE #268: Add driving license class to keycloack

### LINKED ISSUE #269: Add user roles
Add user roles (like tbz) so it's possible to e.g. show who can drive a vehicle.

### LINKED ISSUE #301: Add new user attribute: Status
We need another Keycloak property for the current user status.
The status can be:
"Verfügbar", "Nicht Verfügbar", "Im Einsatz"

Mohammad96Assaf:
close #301
close #269
close #268

### DIFF
--- a/internal/entities/role.go
+++ b/internal/entities/role.go
@@ -1,5 +1,38 @@
 package entities
 
+type UserRole string
+
+const (
+	UserRoleAdmin      UserRole = "Admin"
+	UserRoleDriver     UserRole = "Driver"
+	UserRoleEngineer   UserRole = "Engineer"
+	UserRoleHelper     UserRole = "Helper"
+	UserRoleManagement UserRole = "Management"
+	UserRoleStudent    UserRole = "Student"
+	UserRoleUnknown    UserRole = "unknown"
+)
+
 type Role struct {
-	Name string
+	ID          int32
+	Name        UserRole
+	Description string
+}
+
+func ParseUserRole(role string) UserRole {
+	switch role {
+	case string(UserRoleAdmin):
+		return UserRoleAdmin
+	case string(UserRoleDriver):
+		return UserRoleDriver
+	case string(UserRoleEngineer):
+		return UserRoleEngineer
+	case string(UserRoleHelper):
+		return UserRoleHelper
+	case string(UserRoleManagement):
+		return UserRoleManagement
+	case string(UserRoleStudent):
+		return UserRoleStudent
+	default:
+		return UserRoleUnknown
+	}
 }
--- a/internal/entities/role.go
+++ b/internal/entities/role.go
@@ -3,35 +3,25 @@ package entities
 type UserRole string
 
 const (
-	UserRoleAdmin      UserRole = "Admin"
-	UserRoleDriver     UserRole = "Driver"
-	UserRoleEngineer   UserRole = "Engineer"
-	UserRoleHelper     UserRole = "Helper"
-	UserRoleManagement UserRole = "Management"
-	UserRoleStudent    UserRole = "Student"
-	UserRoleUnknown    UserRole = "unknown"
+	UserRoleTbz               UserRole = "tbz"
+	UserRoleGreenEcolution    UserRole = "green-ecolution"
+	UserRoleSmarteGrenzregion UserRole = "smarte-grenzregion"
+	UserRoleUnknown           UserRole = "unknown"
 )
 
 type Role struct {
-	ID          int32
-	Name        UserRole
-	Description string
+	ID   int32
+	Name UserRole
 }
 
 func ParseUserRole(role string) UserRole {
 	switch role {
-	case string(UserRoleAdmin):
-		return UserRoleAdmin
-	case string(UserRoleDriver):
-		return UserRoleDriver
-	case string(UserRoleEngineer):
-		return UserRoleEngineer
-	case string(UserRoleHelper):
-		return UserRoleHelper
-	case string(UserRoleManagement):
-		return UserRoleManagement
-	case string(UserRoleStudent):
-		return UserRoleStudent
+	case string(UserRoleTbz):
+		return UserRoleTbz
+	case string(UserRoleGreenEcolution):
+		return UserRoleGreenEcolution
+	case string(UserRoleSmarteGrenzregion):
+		return UserRoleSmarteGrenzregion
 	default:
 		return UserRoleUnknown
 	}
--- a/internal/entities/role.go
+++ b/internal/entities/role.go
@@ -14,15 +14,15 @@ type Role struct {
 	Name UserRole
 }
 
-func ParseUserRole(role string) UserRole {
-	switch role {
+func (r *Role) SetName(roleName string) {
+	switch roleName {
 	case string(UserRoleTbz):
-		return UserRoleTbz
+		r.Name = UserRoleTbz
 	case string(UserRoleGreenEcolution):
-		return UserRoleGreenEcolution
+		r.Name = UserRoleGreenEcolution
 	case string(UserRoleSmarteGrenzregion):
-		return UserRoleSmarteGrenzregion
+		r.Name = UserRoleSmarteGrenzregion
 	default:
-		return UserRoleUnknown
+		r.Name = UserRoleUnknown
 	}
 }
--- a/internal/server/http/entities/role.go
+++ b/internal/server/http/entities/role.go
@@ -3,19 +3,15 @@ package entities
 type UserRole string // @Name UserRole
 
 const (
-	UserRoleAdmin      UserRole = "Admin"
-	UserRoleDriver     UserRole = "Driver"
-	UserRoleEngineer   UserRole = "Engineer"
-	UserRoleHelper     UserRole = "Helper"
-	UserRoleManagement UserRole = "Management"
-	UserRoleStudent    UserRole = "Student"
-	UserRoleUnknown    UserRole = "unknown"
+	UserRoleTbz               UserRole = "tbz"
+	UserRoleGreenEcolution    UserRole = "green-ecolution"
+	UserRoleSmarteGrenzregion UserRole = "smarte-grenzregion"
+	UserRoleUnknown           UserRole = "unknown"
 )
 
 type RoleResponse struct {
-	ID          int32  `json:"id"`
-	Name        string `json:"name"`
-	Description string `json:"description"`
+	ID   int32  `json:"id"`
+	Name string `json:"name"`
 } // @Name Role
 
 type RoleListResponse struct {
--- a/internal/server/http/entities/role.go
+++ b/internal/server/http/entities/role.go
@@ -1,7 +1,19 @@
 package entities
 
+type UserRole string // @Name UserRole
+
+const (
+	UserRoleAdmin      UserRole = "Admin"
+	UserRoleDriver     UserRole = "Driver"
+	UserRoleEngineer   UserRole = "Engineer"
+	UserRoleHelper     UserRole = "Helper"
+	UserRoleManagement UserRole = "Management"
+	UserRoleStudent    UserRole = "Student"
+	UserRoleUnknown    UserRole = "unknown"
+)
+
 type RoleResponse struct {
-	ID          string `json:"id"`
+	ID          int32  `json:"id"`
 	Name        string `json:"name"`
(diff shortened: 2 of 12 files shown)


internal/entities/role.go
```diff
@@ -1,5 +1,38 @@
 package entities
 
+type UserRole string
+
+const (
+	UserRoleAdmin      UserRole = "Admin"
```
doriengr, line 6: I am not sure if those roles are helpful. Currently we need those roles to check if a person is from the tbz, digital agency or from the university.

The background is that we only want to link the person who is from the TBZ in the frontend to a watering plan.

internal/server/http/entities/role.go
```diff
@@ -1,7 +1,19 @@
...
+	UserRoleManagement UserRole = "Management"
+	UserRoleStudent    UserRole = "Student"
+	UserRoleUnknown    UserRole = "unknown"
+)
+
 type RoleResponse struct {
-	ID          string `json:"id"`
+	ID          int32  `json:"id"`
 	Name        string `json:"name"`
 	Description string `json:"description"`
```
doriengr, line 18: Do the roles have a description in keycloak?

internal/server/http/handler/v1/user/handler.go
```diff
@@ -187,9 +183,63 @@ func parseURL(rawURL string) (*url.URL, error) {
...
+// @Produce		json
+// @Success		200		{object}	entities.UserListResponse
+// @Failure		400		{object}	HTTPError
+// @Failure		500		{object}	HTTPError
+// @Param			page	query		string	false	"Page"
+// @Param			limit	query		string	false	"Limit"
+// @Param			limit	query		string	false	"Role"
+// @Router			/v1/user [get]
+// @Security		Keycloak
+func GetUsersByRole(svc service.AuthService) fiber.Handler {
```
doriengr, line 217: Cedrik left on my PR to the vehicles a comment that it would be better if we use for this kind of stuff parameter and not a new endpoint. 

so GetAll but use something like `users?role=tbz`

doriengr (changes requested):

internal/server/http/entities/role.go
xk0niR: @doriengr Would be a description "worldwide", e.g. user-role "engineer" is the same for every person?

I am mostly sure they don´t have a description within keycloak. While inspecting I noticed a user inherits attributes from "Role mapping" (e.g. normal users are mapped to 'green-ecolution-dev'). If you then take a look at "User profile" of 'green-ecolution-dev' you can see (custom) attributes like

- user_roles ${profile.attributes.user_roles}

Inside keycloak there is not restriction, e.g. a user_role can be set to 'firefighter'.

I guess there are two options:
1. Restrict within keycloak possible user_role (if possible) and save description within keycloak
2. Let users/admins change roles while using frontend. We then can easy restrict possible options and set descriptions which would make us less dependent on keycloak.

Can @choffmann and @Mohammad96Assaf comment on this?

xk0niR: Description of roles are not needed (also not really possible within keycloak)

Mohammad96Assaf: Thank you, Konstantin, for clarifying it and restricting the roles within Keycloak 👍

doriengr:
The tests are not running somehow. Maybe because of the merge conflicts?

Mohammad96Assaf:
I added the user status attribute to this ticket, and I removed the open branch related to the status issue.

internal/server/http/handler/v1/user/handler.go
```diff
@@ -218,6 +218,47 @@ func GetAllUsers(svc service.AuthService) fiber.Handler {
 	}
 }
 
+// @Summary		Get users by role
```
doriengr, line 221: the id is missing. With the id we can give the endpoint a name. 

`// @Id get-users-by-role` or something like that maybe?

internal/server/http/handler/v1/user/handler.go
```diff
@@ -218,6 +218,47 @@ func GetAllUsers(svc service.AuthService) fiber.Handler {
...
+		ctx := c.Context()
+
+		role := strings.Clone(c.Params("Role"))
+		if role == "" {
+			return errorhandler.HandleError(service.NewError(service.BadRequest, "invalid role format"))
+		}
+
+		userRole := domain.ParseUserRole(role)
+		if userRole == domain.UserRoleUnknown {
+			return errorhandler.HandleError(service.NewError(service.BadRequest, "invalid role name"))
```
doriengr, line 244: Why not get all users with the role unknown?
The »invalid role« error should be happening if the role parsing did not work or if the request has an invalid string that is not a role.

internal/server/http/handler/v1/user/handler.go
```diff
@@ -218,6 +218,47 @@ func GetAllUsers(svc service.AuthService) fiber.Handler {
...
+			return errorhandler.HandleError(service.NewError(service.InternalError, errors.Wrap(err, "failed to get users by role").Error()))
+		}
+		response := make([]entities.UserResponse, len(users))
+
+		for i, user := range users {
+			userResponse := userMapper.FromResponse(user)
+			response[i] = *userResponse
+		}
+
+		return c.Status(fiber.StatusOK).JSON(response)
```
doriengr, line 258: Let'sreturn it like this: 

```
		return c.Status(fiber.StatusOK).JSON(entities.UserListResponse{
			Data:       data,
			Pagination: entities.Pagination{}, // TODO: Handle pagination
		})
```

I know, we do not have a pagination (yet), but then we can simply implement the pagination later on and we do not have to update the frontend. 👍

internal/server/http/handler/v1/user/handler_test.go
```diff
@@ -683,6 +683,130 @@ func TestGetAllUsers(t *testing.T) {
...
+		app := fiber.New()
+		mockAuthService := serviceMock.NewMockAuthService(t)
+		app.Get("/v1/user/role/:role", GetUsersByRole(mockAuthService))
+
+		mockUUID1 := uuid.New()
+		mockUUID2 := uuid.New()
+
+		users := []*domain.User{
+			{
+				ID:          mockUUID1,
```
doriengr, line 698: Why not directly`ID: uuid.New()`?

internal/server/http/handler/v1/user/routes.go
```diff
@@ -8,6 +8,7 @@ import (
 func RegisterRoutes(r fiber.Router, svc service.AuthService) {
 	r.Post("/", Register(svc))
 	r.Get("/", GetAllUsers(svc))
+	r.Get("/role/:role", GetUsersByRole(svc))
```
doriengr, line 11: Do not forget to implement a test in `routes_test` for this endpoint.

internal/service/domain/auth/user.go
```diff
@@ -84,3 +84,22 @@ func (s *AuthService) GetByIDs(ctx context.Context, ids []string) ([]*domain.Use
 
 	return users, nil
 }
+
+func (s *AuthService) GetAllByRole(ctx context.Context, role domain.Role) ([]*domain.User, error) {
+	users, err := s.GetAll(ctx)
```
doriengr, line 89: Is it not `s.userRepo.GetAll(ctx)`?

I am not sure but in the keycloak client there a functions like `GetUsersByRoleName` and stuff like that. Maybe you can use this in the userRepo and then you do not have to use a filter and loop over all the users.

But maybe I am wrong. I am not sure.

doriengr (changes requested):

internal/service/domain/auth/user.go
Mohammad96Assaf: There is a difference between the roles inside Keycloak and the "custom user-role attribute" that we defined. The roles in Keycloak cannot be restricted, whereas the "custom user-role attribute" we defined is limited to three specific values: tbz, green-ecolution, and smarte-grenzregion. To enforce this restriction, we added the custom attribute inside Keycloak.

So, to use the attribute in the query, I called getAllUsers and filtered the results according to the role name.

internal/server/http/entities/role.go
```diff
@@ -1,9 +1,17 @@
 package entities
 
+type UserRole string // @Name UserRole
+
+const (
+	UserRoleTbz               UserRole = "tbz"
+	UserRoleGreenEcolution    UserRole = "green-ecolution"
+	UserRoleSmarteGrenzregion UserRole = "smarte-grenzregion"
+	UserRoleUnknown           UserRole = "unknown"
```
doriengr, line 9: If currently a user has no role the returned value is empty and not »unknown«. In keycloak there is not role »unknown« possible to select. Either we map the keycloak user to the current user like that: if a user has no role, the role is set to »unknown« in our system. Same logik for the status or we just delete the user role unknown.

 

image

Mohammad96Assaf: This happens because the attributes were newly defined in Keycloak. Once we select a value, we cannot leave it empty again. 
For example, with the status attribute, once it is selected, the second time we select it, we are forced to switch between its two values "absent or available".
I will open Keycloak and select values for status and roles for the first time, and after that, we will not see empty values again.

doriengr: But can we define »Unkown« as a default in our system then? :)

internal/entities/role.go
```diff
@@ -1,5 +1,28 @@
...
+	UserRoleUnknown           UserRole = "unknown"
+)
+
 type Role struct {
-	Name string
+	ID   int32
+	Name UserRole
+}
+
+func ParseUserRole(role string) UserRole {
```
choffmann, line 17: Please make this a member function of `Role`

internal/server/http/entities/role.go
Mohammad96Assaf: We cannot define the roles as "unknown" because it is a list attribute. However, it will now be represented as an empty list in the JSON. 
The status will be set to "unknown" if no status is defined.

```
"roles": [],
"driving_license": "B",
"status": "unknown"
```

doriengr: Sorry, I do not get the problem. The Type »UserStatusUnknown« needs to be defined, because we need it in the frontend!

internal/server/http/entities/user.go
```diff
@@ -4,6 +4,13 @@ import (
 	"time"
 )
 
+type UserStatus string // @Name UserStatus
+
+const (
```
doriengr, line 9:
```
const (
	UserStatusAvailable     UserStatus = "available"
	UserStatusAbsent        UserStatus = "absent"
        UserStatusUnknown    UserStatus = "unknown"
)
```

internal/storage/auth/keycloak/user.go
```diff
@@ -145,17 +146,37 @@ func keyCloakUserToUser(user *gocloak.User) (*entities.User, error) {
...
 		}
+
+		if val, ok := (*user.Attributes)["driving_license_class"]; ok && len(val) > 0 {
+			drivingLicenseClass = val[0]
+		}
+
+		if val, ok := (*user.Attributes)["user_roles"]; ok && len(val) > 0 {
+			userRoles = val
+		}
+
```
doriengr, line 157: Please check if the status is one one of the »UserStatus« types we defined in our system and if not or if the status is empty — return the »UserStatusUnknown«!

doriengr (changes requested):

choffmann (approved):
My requested changes are approved :+1:

internal/server/http/entities/user.go
Mohammad96Assaf: Did you check the last changes?!

internal/storage/auth/keycloak/user.go
Mohammad96Assaf: Did you check the last changes?!

doriengr: Oh sorry, but there is still a mistake:

```
const (
	UserStatusAvailable SensorStatus = "available"
	UserStatusAbsent    SensorStatus = "absent"
	UserStatusUnknown   SensorStatus = "unknown"
)
```

This is not the SensorStatus.

doriengr: Please change this. Then you can merge 👍

doriengr (approved):

PR #299, opened 2024-12-28
Feature/schedular for planned watering plans

### LINKED ISSUE #291: Schedular to check if watering plans are not competed
We need some kind of logic to check if planned watering plans are finished on their planned date - if not, the status of the watering plan should be set to »not competed«

doriengr:
Close #291

### DIFF
--- a/internal/service/domain/watering_plan/status_schedular.go
+++ b/internal/service/domain/watering_plan/status_schedular.go
@@ -1,0 +1,68 @@
+package wateringplan
+
+import (
+	"context"
+	"fmt"
+	"log/slog"
+	"time"
+
+	"github.com/green-ecolution/green-ecolution-backend/internal/entities"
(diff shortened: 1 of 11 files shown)


internal/service/domain/watering_plan/status_schedular.go
```diff
@@ -0,0 +1,68 @@
...
+					"error", err.Error(),
+				)
+			} else {
+				slog.Info("Watering plan marked as not competed", "id", wateringPlan.ID)
+			}
+		}
+	}
+
+	return nil
+}
```
choffmann, line 68: Should we merge this status scheduler for the watering schedule and the sensor status schedule? Or make it more generic? I also commented on this in Mohammad's PR, but it was not considered. 

I'm thinking of some worker struct that can be started separately and given a function to do the work stuff

doriengr:
I am closing this PR and I am going to open a new one. That way it's easier to use the new generic scheduler.

PR #306, opened 2024-12-30
Feature/get user by ids

### LINKED ISSUE #305: Fix test case for TestKeyCloakUserRepo_GetAll

### LINKED ISSUE #307: Get user by ids in storage, service and server

doriengr:
Close #305 
Close #307 

We need a way to get several users back at the same time with their ID so that we can find out which users are linked in one watering plan (in the frontend).


xk0niR (approved):
Well done!!! :100:

PR #324, opened 2025-01-05
Feature/update watering plan event

doriengr:


internal/worker/subscriber/subscriber.go
```diff
@@ -88,3 +88,28 @@ func (s *CreateSensorDataSubscriber) HandleEvent(ctx context.Context, e entities
...
+
+type UpdateWateringPlanSubscriber struct {
+	tcSvc service.TreeClusterService
+	wpSvc service.WateringPlanService
+}
+
+func NewUpdateWateringPlanSubscriber(tcSvc service.TreeClusterService, wpSvc service.WateringPlanService) *UpdateWateringPlanSubscriber {
+	return &UpdateWateringPlanSubscriber{
+		tcSvc: tcSvc,
+		wpSvc: wpSvc,
```
choffmann, line 100: Do we need the wateringplan service here? You do not use it in the subscriber

internal/service/domain/treecluster/handle_watering_plan_update_test.go
choffmann, line 1: If you are implementing the publish treecluster update event, make sure you test the event publishing here as well

choffmann (changes requested):

PR #340, opened 2025-01-10
fix: tree cluster update on empty trees

### LINKED ISSUE #336: Creating a treecluster throws error
Sometimes error code 405 or 502
image

doriengr:
Close #339 
Close #336 

I am not sure if this is a nice way to fix this error. Please check if there are other possibilities and if this is clean enough …

### DIFF
--- a/.docker/docker-compose.infra.yaml
+++ b/.docker/docker-compose.infra.yaml
@@ -30,7 +30,7 @@ services:
     ports:
       - "8080:8082"
       - "9001:9001"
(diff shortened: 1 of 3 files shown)


choffmann (approved):
Ahh yes of course, if a cluster was created without trees, then the cluster doesn't have lat and long attributes, which was needed in the transaction and returend an error. I just added some logging and do not return an error if a region was not found by the coordinates. So the region is then nil, but the cluster has coordinate values.

Good job :+1:

PR #351, opened 2025-01-11
feat: return route metadata on generating route

choffmann:
Return information about the generated route, such as distance, time and refills


internal/server/http/entities/watering_plan.go
```diff
@@ -33,6 +33,8 @@ type WateringPlanResponse struct {
 	CancellationNote   string                       `json:"cancellation_note"`
 	Evaluation         []*EvaluationValue           `json:"evaluation"`
 	GpxURL             string                       `json:"gpx_url"`
+	Duration           *float64                     `json:"duration"`
+	RefillCount        int32                        `json:"refill_count"`
```
doriengr, line 37: Why not everywhere `Refills` or `RefillCount`?

doriengr (approved):
LGTM 🔥

PR #353, opened 2025-01-11
chor: validate required user attributes

### LINKED ISSUE #346: Handle user attributes if they are nil (e.g., user.Username, user.FirstName, user.LastName, user.Email)

Mohammad96Assaf:
close #346


internal/storage/auth/keycloak/user.go
```diff
@@ -186,6 +185,26 @@ func keyCloakUserToUser(user *gocloak.User) (*entities.User, error) {
...
+		userRole.SetName(roleName)
+		roles = append(roles, userRole)
+	}
+	if roles == nil {
+		roles = []entities.Role{}
+	}
+	return roles
+}
+
+func validateRequiredAttributes(user *gocloak.User) error {
```
doriengr, line 201: I do not know what the naming of it is but should these functions not be:

`func (r *UserRepository) .....`

doriengr: Yes. I did not know that its called a member function.

Mohammad96Assaf: It is a private helper function used within the `keyCloakUserToUser` function. I created it because the original function had become too large, causing a `linter `error. Moreover, we cannot make it a member function since `keyCloakUserToUser` itself is not a member function.

doriengr: And why is `keyCloakUserToUser` not a member function then?

Mohammad96Assaf: Because it is a `private helper function`, and we do not use it outside the file `user.go`. Also, we do not need to access it through a member of the `UserRepository` struct.
I mean: If the function doesn’t need to access or modify any data from the `UserRepository `struct, there’s no reason to make it a member function. :)

doriengr (approved):

PR #362, opened 2025-01-17
feat: check for correct user roles and driving licenses

### LINKED ISSUE #333: Check correct driving license of users linked to a watering plan

### LINKED ISSUE #363: Check correct user role of users linked to a watering plan

doriengr:
Close #363 
Close #333 

- [x] refactor user roles
- [x] add »CE« to driving licenses
- [x] make driving licenses an array
- [x] check correct user role on create and update of watering plans
- [x] check correct driving licenses on create and update

**Todo:**
- delete old driving license class value from keycloak realm
- update JWT token
- update frontend to use correct values


choffmann:
@doriengr I rebased develop in your PR and change some structure of the functions. For example, I changed the `fetchVehicle()` function. This function has no use because it's called as a function and returns the error if there is one. But when you call this function, you also have to handle the error. You are abstracting a function with no benefit and at the cost of detailed logging. 

So I changed this:
```go 
transporter, err := w.fetchVehicle(ctx, *createWp.TransporterID)
if err != nil {
	return nil, service.MapError(ctx, err, service.ErrorLogEntityNotFound)
}
```

To this:
```go
transporter, err := w.vehicleRepo.GetByID(ctx, *createWp.TransporterID)
if err != nil {
	log.Debug("failed to get transporter by id", "error", err, "transporter_id", *createWp.TransporterID)
	return nil, service.MapError(ctx, err, service.ErrorLogEntityNotFound)
}
```

I hope you see what I mean. It's a good practice in other languages but I notice in Go this it's get me more confused (one thing I find annoying about Go). I just wanted to mention it briefly. I will do the review tomorrow :+1:

choffmann (approved):

PR #378, opened 2025-01-20
Feature/add tree cluster pagination

### LINKED ISSUE #370: Implement pagination for treecluster handler

doriengr:
Close #370


internal/server/http/handler/v1/treecluster/handler.go
```diff
@@ -28,12 +28,30 @@ var (
...
+		if limitParam != "" {
+			limitParam, err := strconv.Atoi(limitParam)
+			if err != nil || (limitParam != -1 && limitParam <= 0) {
+				err := service.NewError(service.BadRequest, "invalid limit format")
+				return errorhandler.HandleError(err)
+			}
+			limit = int32(limitParam)
+		}
+
+		domainData, totalCount, err := svc.GetAll(ctx, page, limit)
```
choffmann, line 54: I would love to see this in the middleware. So we can use it with every entity in our app. Here is a small blog post on how to implement pagination in the middleware. This is an example for gin, but you can easily translate it to fibers middleware.

https://articles.wesionary.team/make-pagination-easy-in-golang-using-pagination-middleware-and-gorm-scope-a5f6eb3bebaa

internal/server/http/handler/v1/treecluster/handler.go
```diff
@@ -43,9 +61,22 @@ func GetAllTreeClusters(svc service.TreeClusterService) fiber.Handler {
...
+
+			pagination = &entities.Pagination{
+				Total:       totalCount,
+				CurrentPage: page,
+				TotalPages:  totalPages,
+				NextPage:    nextPage,
+				PrevPage:    prevPage,
+			}
+		}
+
```
choffmann, line 76: With the fiber middleware, this will be put into a context and can be used everywhere in our app, that has this context

internal/storage/postgres/queries/treecluster.sql
```diff
@@ -1,5 +1,10 @@
 -- name: GetAllTreeClusters :many
-SELECT * FROM tree_clusters ORDER BY name ASC;
+SELECT * FROM tree_clusters 
+ORDER BY name ASC
+LIMIT $1 OFFSET $2;
```
choffmann, line 4: :+1:

choffmann (changes requested):

internal/server/http/handler/v1/treecluster/handler.go
```diff
@@ -33,23 +35,8 @@ func GetAllTreeClusters(svc service.TreeClusterService) fiber.Handler {
...
-			if err != nil || (limitParam != -1 && limitParam <= 0) {
-				err := service.NewError(service.BadRequest, "invalid limit format")
-				return errorhandler.HandleError(err)
-			}
-			limit = int32(limitParam)
-		}
+		page := c.Locals(middleware.Page).(int32)
+        limit := c.Locals(middleware.Limit).(int32)
 
 		domainData, totalCount, err := svc.GetAll(ctx, page, limit)
```
choffmann, line 41: The functions, in this case `GetAll`, does not need to take `page` and `limit` as parameters. These attributes are set in the context

choffmann (commented):
Looks good so far :+1:

internal/storage/postgres/treecluster/get.go
```diff
@@ -7,26 +7,55 @@ import (
...
+	if page <= 0 || (limit != -1 && limit <= 0) {
+		log.Debug("invalid pagination values")
+		return nil, 0, r.store.MapError(storage.ErrPaginationValueInvalid, sqlc.TreeCluster{})
+	}
+
+	totalCount, err := r.store.GetAllTreeClustersCount(ctx)
+	if err != nil {
+		log.Debug("failed to get total tree cluster count in db")
+		return nil, 0, r.store.MapError(err, sqlc.TreeCluster{})
+	}
```
choffmann, line 34: We may be able to return early at this point if we know there are no records in the query (`totalCount == 0`)

internal/storage/postgres/treecluster/get.go
```diff
@@ -7,26 +7,55 @@ import (
...
+
+	if page <= 0 || (limit != -1 && limit <= 0) {
+		log.Debug("invalid pagination values")
+		return nil, 0, r.store.MapError(storage.ErrPaginationValueInvalid, sqlc.TreeCluster{})
+	}
+
+	totalCount, err := r.store.GetAllTreeClustersCount(ctx)
+	if err != nil {
+		log.Debug("failed to get total tree cluster count in db")
```
choffmann, line 32: Please add the error to the log

```suggestion 
		log.Debug("failed to get total tree cluster count in db", "error", err)
```

internal/utils/pagination.go
```diff
@@ -0,0 +1,54 @@
+package utils
```
choffmann, line 1: What do you think about moving this to a separate `pagination` package?

internal/storage/postgres/treecluster/get.go
```diff
@@ -7,26 +7,55 @@ import (
...
 
-func (r *TreeClusterRepository) GetAll(ctx context.Context) ([]*entities.TreeCluster, error) {
+func (r *TreeClusterRepository) GetAll(ctx context.Context) ([]*entities.TreeCluster, int64, error) {
 	log := logger.GetLogger(ctx)
-	rows, err := r.store.GetAllTreeClusters(ctx)
+	page, pageOk := ctx.Value("page").(int32)
+	limit, limitOK := ctx.Value("limit").(int32)
+
+	if !pageOk || !limitOK {
+		log.Debug("pagination values not found in context")
```
choffmann, line 21: Maybe print the raw values inside the context
```suggestion
		log.Debug("pagination values not found in context", "page", ctx.Value("page"), "limit", ctx.Value("limit")
```

internal/storage/postgres/treecluster/get.go
```diff
@@ -7,26 +7,55 @@ import (
...
+		return nil, 0, r.store.MapError(storage.ErrPaginationValueInvalid, sqlc.TreeCluster{})
+	}
+
+	if page <= 0 || (limit != -1 && limit <= 0) {
+		log.Debug("invalid pagination values")
```
choffmann, line 26: Can you put more information to this log. Why are the values invalid pagination values?

internal/storage/postgres/treecluster/get.go
```diff
@@ -7,26 +7,55 @@ import (
...
+		page = 1
+	}
+
+	rows, err := r.store.GetAllTreeClusters(ctx, &sqlc.GetAllTreeClustersParams{
+		Limit:  limit,
+		Offset: (page - 1) * limit,
+	})
+
 	if err != nil {
 		log.Debug("failed to get tree clusters in db")
```
choffmann, line 47: This may be my mistake, but can you also put the error message in the debug log?

```suggestion
		log.Debug("failed to get tree clusters in db", "error", err)
```

choffmann (changes requested):

internal/storage/postgres/treecluster/get.go
```diff
@@ -7,26 +7,55 @@ import (
...
+
+	if !pageOk || !limitOK {
+		log.Debug("pagination values not found in context")
+		return nil, 0, r.store.MapError(storage.ErrPaginationValueInvalid, sqlc.TreeCluster{})
+	}
+
+	if page <= 0 || (limit != -1 && limit <= 0) {
+		log.Debug("invalid pagination values")
+		return nil, 0, r.store.MapError(storage.ErrPaginationValueInvalid, sqlc.TreeCluster{})
+	}
```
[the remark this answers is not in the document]
doriengr: Where would you put the pagination stuff? Where is the pagination package located?

internal/storage/postgres/treecluster/get.go
choffmann: Sure, you can test it. If there is no value, the value is `nil`, otherwise an actual value which the logger will try to display as a string.

This line will produce this output:

```go
slog.Info("test to get not existing value in ctx", "ctx_value", ctx.Value("not_existing"))
```

```
[Jan 26 16:32:13] INFO main.go:84 test to get not existing value in ctx ctx_value="<nil>"
```

internal/storage/postgres/treecluster/get.go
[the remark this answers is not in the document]
choffmann: :man_shrugging: Maybe in `internal/utils/pagination` or `internal/pagination` or `internal/service/domain/pagination`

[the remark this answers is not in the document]
doriengr: Good idea 🔥

internal/server/http/handler/v1/treecluster/handler.go
```diff
@@ -28,12 +29,12 @@ var (
...
 // @Param			page	query	string	false	"Page"
 // @Param			limit	query	string	false	"Limit"
-// @Param			status	query	string	false	"Status"
 // @Security		Keycloak
 func GetAllTreeClusters(svc service.TreeClusterService) fiber.Handler {
 	return func(c *fiber.Ctx) error {
 		ctx := c.Context()
-		domainData, err := svc.GetAll(ctx)
+
+		domainData, totalCount, err := svc.GetAll(ctx)
```
choffmann, line 37: Is `totalCount` not the same as `len(domainData)`? Is there a reason why you return a seperate value? So I would suggest this


```suggestion
		domainData, err := svc.GetAll(ctx)
		totalCount := len(domainData)
```

internal/storage/postgres/treecluster/get.go
choffmann: Did you forget to push this? :sweat_smile: There is still the error attribute missing in the log

internal/storage/postgres/treecluster/get.go
```diff
@@ -8,25 +8,50 @@ import (
 	"github.com/green-ecolution/green-ecolution-backend/internal/entities"
 	"github.com/green-ecolution/green-ecolution-backend/internal/logger"
 	sqlc "github.com/green-ecolution/green-ecolution-backend/internal/storage/postgres/_sqlc"
+	"github.com/green-ecolution/green-ecolution-backend/internal/utils/pagination"
 	"github.com/twpayne/go-geos"
 )
 
-func (r *TreeClusterRepository) GetAll(ctx context.Context) ([]*entities.TreeCluster, error) {
+func (r *TreeClusterRepository) GetAll(ctx context.Context) ([]*entities.TreeCluster, int64, error) {
```
choffmann, line 15: Same question here, is there a reason why you return the size seperately instead of `len(...)`?

choffmann (changes requested):

internal/server/http/handler/v1/treecluster/handler.go
doriengr: The domainData is the paginated treeCluster data isn't it? That's not the `totalCount` of all tree cluster

choffmann: True, sorry my Bad 🫣

choffmann (approved):

PR #381, opened 2025-01-21
feature: add generic scheduler

### LINKED ISSUE #365: Create a generic infrastructure for schedulars

Mohammad96Assaf:
close #365


internal/utils/utils.go
```diff
@@ -53,3 +56,17 @@ func UUIDToPGUUID(userID uuid.UUID) pgtype.UUID {
 		Valid: true,
 	}
 }
+
+func Scheduler[T any](ctx context.Context, interval time.Duration, process func(ctx context.Context) T) {
```
choffmann, line 60: Why you use the generic type `[T any]` when you dont do something with it?

choffmann, line 60: As it is now, the ticker would only start after the first interval. So, for example, with an interval of one hour, the ticker would only execute the function for the first time after one hour, is that what we want?

choffmann (changes requested):

internal/utils/utils.go
Mohammad96Assaf: The generic [T any] ensures Scheduler can handle functions with any return type, making it flexible.
In the sensor case, we don't use the return type, but it allows functions returning error, string, etc.
Without it, Scheduler would only work with functions that return nothing.

internal/service/domain/sensor/sensor.go
```diff
@@ -129,8 +127,25 @@ func (s *SensorService) Delete(ctx context.Context, id string) error {
...
-	s.StatusUpdater.RunStatusUpdater(ctx, interval)
+func (s *SensorService) UpdateStatuses(ctx context.Context) error {
+	sensors, err := s.sensorRepo.GetAll(ctx)
+	if err != nil {
+		return err
+	}
+
+	cutoffTime := time.Now().Add(-72 * time.Hour) // 3 days ago
+	for _, sens := range sensors {
+		if sens.UpdatedAt.Before(cutoffTime) {
```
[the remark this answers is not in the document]
Mohammad96Assaf: Yes, you are right, not always. Therefore, I removed the `if` statement from the `getSensor` function in `internal/service/domain/sensor/mqtt.go` to ensure the sensor is updated whenever we receive data. Now, the sensor will always be updated when we receive data from `TTN`. I hope this is a good solution.

internal/utils/utils.go
Mohammad96Assaf: Yes, it happens only the first time the ticker waits. However, we can call the function outside the loop to execute it immediately before waiting for the ticker.

choffmann: Yes, but this is not used, is it? A function should adapt to the scheduler, not the scheduler to the function. What I can imagine, if at all, is that a function returns an error, but then it must not be generic. And if the scheduler gets an error, what should happen? Should the scheduler try a certain number of times and then give up?

internal/worker/Scheduler.go
```diff
@@ -0,0 +1,35 @@
...
+	ticker := time.NewTicker(interval)
+	defer ticker.Stop()
+
+	if ctx.Err() == nil {
+		err := process(ctx)
+		if err != nil {
+			slog.Error("Error during initial process execution", "error", err)
+		}
+	} else {
+		slog.Info("Stopping scheduler before first execution due to canceled context")
```
choffmann, line 19: please get the logger from the context and keep the log message in lowercase

```go
log := logger.GetLogger(ctx)
```

choffmann: Is it really necessary to display this log, what defence does it give the user?  Maybe turn this into an debug log?

internal/worker/Scheduler.go
```diff
@@ -0,0 +1,35 @@
...
+)
+
+func Scheduler(ctx context.Context, interval time.Duration, process func(ctx context.Context) error) {
+	ticker := time.NewTicker(interval)
+	defer ticker.Stop()
+
+	if ctx.Err() == nil {
+		err := process(ctx)
+		if err != nil {
+			slog.Error("Error during initial process execution", "error", err)
```
choffmann, line 16: please get the logger from the context and keep the log message in lowercase

```go
log := logger.GetLogger(ctx)
```

internal/worker/Scheduler.go
```diff
@@ -0,0 +1,35 @@
...
+		slog.Info("Stopping scheduler before first execution due to canceled context")
+		return
+	}
+
+	for {
+		select {
+		case <-ticker.C:
+			err := process(ctx)
+			if err != nil {
+				slog.Error("Error during process execution", "error", err)
```
choffmann, line 28: please get the logger from the context and keep the log message in lowercase

```go
log := logger.GetLogger(ctx)
```

internal/worker/Scheduler.go
```diff
@@ -0,0 +1,35 @@
...
+			}
+		case <-ctx.Done():
+			slog.Info("Stopping scheduler")
```
choffmann, line 31: please get the logger from the context and keep the log message in lowercase

```go
log := logger.GetLogger(ctx)
```

internal/worker/Scheduler.go
choffmann, line 1: Please keep a consistent naming convention. This file should be in lower case

internal/worker/Scheduler_test.go
choffmann, line 1: Please keep a consistent naming convention. This file should be in lower case

choffmann (changes requested):

internal/worker/scheduler.go
```diff
@@ -0,0 +1,38 @@
...
+		log.Debug("stopping scheduler before first execution due to canceled context")
+		return
+	}
+
+	for {
+		select {
+		case <-ticker.C:
+			err := process(ctx)
+			if err != nil {
+				slog.Error("error during process execution", "error", err)
```
choffmann, line 31:
```suggestion
				log.Error("error during process execution", "error", err)
```

internal/worker/scheduler.go
```diff
@@ -0,0 +1,38 @@
...
+			}
+		case <-ctx.Done():
+			slog.Debug("stopping scheduler")
```
choffmann, line 34:
```suggestion
			log.Debug("stopping scheduler")
```

internal/service/domain/sensor/sensor.go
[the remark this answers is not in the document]
choffmann: Have you committed the changes? I don't see it and don't think you can use the `UpdatedAt` field on the sensor to determine this

internal/worker/scheduler.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"context"
+	"log/slog"
+	"time"
+
+	"github.com/green-ecolution/green-ecolution-backend/internal/logger"
+)
+
+func Scheduler(ctx context.Context, interval time.Duration, process func(ctx context.Context) error) {
```
choffmann, line 11: I think it would be nicer to have the scheduler work directly instead of calling it outside of this function. Have a look at the scheduler of the tbz baumkataster plugin: https://github.com/green-ecolution/tbz-baumkataster-plugin/blob/develop/worker.go

choffmann (changes requested):

internal/worker/scheduler.go
Mohammad96Assaf: What do mean with "directly"?, Do you mean making the scheduler an object that starts automatically instead of needing an external call?

Something like this?

`scheduler := worker.NewScheduler(ctx, 10*time.Second, myProcessFunction)`

and 

```
func NewScheduler(ctx context.Context, interval time.Duration, process func(ctx context.Context) error) *Scheduler {
    ctx, cancel := context.WithCancel(ctx)

    s := &Scheduler{
        interval: interval,
        process:  process,
        ctx:      ctx,
    }

    go s.run() // Start the scheduler automatically
    return s
}
```

choffmann: No, when the schedular starts, and assuming we have a three hour interval, the function that the schedular is to execute would be executed for the first time after three hours. To avoid this, and to avoid having to run the 'work' function separately, I would suggest that the schedular runs the 'work' function directly, and then runs it again and again at the specified interval. 

In the schedular I implemented for the plugin, you can see what I mean

internal/service/domain/sensor/sensor.go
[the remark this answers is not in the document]
Mohammad96Assaf: I implemented it as discussed. Take a look at the latest version now.

internal/worker/scheduler.go
choffmann: Hmm ja okay. But there is no need to check if `ctx.Err() == nil`. So when you want to keep it simple, I would change it to this:


```go
if err := process(ctx); err != nil {
	log.Error("error during initial process execution", "error", err)
}
```

internal/worker/scheduler.go
```diff
@@ -0,0 +1,35 @@
...
+	} else {
+		log.Debug("stopping scheduler before first execution due to canceled context")
+		return
+	}
+
+	for {
+		select {
+		case <-ticker.C:
+			if err := process(ctx); err != nil {
+				log.Error("error during initial process execution", "error", err)
```
choffmann, line 28: This is not an initial process anymore...

internal/service/domain/sensor/mqtt.go
```diff
@@ -92,7 +92,7 @@ func (s *SensorService) updateSensorCoordsAndStatus(ctx context.Context, payload
 		if err != nil {
 			return nil, err
 		}
-
+		log.Debug("Coordinates and status of sensor have been updated successfully", "sensor_id", updatedSensor.ID)
```
choffmann, line 95: Maybe this should be an Info logging. Also, please be consistent in the log messages by writing everything in lower case

choffmann (changes requested):

internal/worker/scheduler.go
choffmann: One thing that still bothers me is that we now have a scheduler, but it can only do one function. It would be better if we could support a structure as well. Take a look at this: https://gist.github.com/choffmann/d11e8a6111faf498965a8869271ff05f
 
Here I have created an interface `SchedularWork` which contains a function `Do()`. This interface can be implemented by other structs and passed to the schedular. This pattern command used for example in the http.Handler in Go. https://pkg.go.dev/net/http#HandlerFunc. The advantage here is that internal state can be managed in the structure. 

The point is to implement a generic scheduler, but as it is, it is still too ungeneric for me.

Mohammad96Assaf: Your suggestion looks good! I've added it to the implementation and additionally created a struct to make it more flexible, such as using the `Stop()` function when **needed**.
I know we can stop the scheduler directly using the given `context`, but every time we need to stop it, we have to define a new line if it's not already defined:
```
ctx, cancel := context.WithCancel(ctx)
cancel()
```
I think, having a Struct makes it more Flexible.

What’s your opinion on this?

internal/worker/scheduler.go
```diff
@@ -0,0 +1,67 @@
...
+import (
+	"context"
+	"time"
+
+	"github.com/green-ecolution/green-ecolution-backend/internal/logger"
+)
+
+type Scheduler struct {
+	ctx      context.Context
+	cancel   context.CancelFunc
```
choffmann, line 12: > Contexts should not be stored inside a struct type, but instead passed to each function that needs it.

https://go.dev/blog/context-and-structs#introduction

internal/server/http/server.go
```diff
@@ -48,9 +50,8 @@ func (s *Server) Run(ctx context.Context) error {
 		}
 	}()
 
-	go func() {
-		s.services.SensorService.RunStatusUpdater(ctx, 1*time.Hour)
-	}()
+	scheduler := worker.NewScheduler(ctx, 3*time.Hour, s.services.SensorService)
+	go scheduler.Run()
```
choffmann, line 54: Now you have passed the sensor service as a struct and implemented the `Do` function in the service. In this case I would prefer to use the function call. Because now we have a `Do` function inside our service... Perhaps the service has two or more schedulars that will do different things in the future, then this will not work. So please rename the function back to `UpdateStatuses` and passed like so:

```suggestion
	scheduler := worker.NewScheduler(3*time.Hour, worker.SchedulerFunc(s.services.SensorService.UpdateStatuses))
	go scheduler.Run(ctx)
```

choffmann (changes requested):
Yes, I understand what you're trying to do, and in general the idea isn't bad either. It's good that the scheduler is now a struct :+1: but... It is a Go Convention that context must not be saved in the struct https://go.dev/blog/context-and-structs#introduction. As you said yourself, the caller should create the `context.WithCancel` itself and call the cancel function.

Mohammad96Assaf:
Thanks for the good information! I’ve applied your suggestions 👍 
So now, the scheduler is initialized like this: 
```
scheduler := worker.NewScheduler(3*time.Hour, worker.SchedulerFunc(s.services.SensorService.UpdateStatuses))
go scheduler.Run(ctx)
```

PR #382, opened 2025-01-21
feat: implement transactions for tree repo

### LINKED ISSUE #366: Implement transactions for tree repository

Mohammad96Assaf:
close #366
ref #85


internal/service/domain/tree/tree.go
```diff
@@ -121,32 +124,43 @@ func (s *TreeService) Create(ctx context.Context, treeCreate *entities.TreeCreat
 		return nil, service.NewError(service.BadRequest, errors.Wrap(err, "validation error").Error())
 	}
 
-	fn := make([]entities.EntityFunc[entities.Tree], 0)
+	var treeClusterID *entities.TreeCluster
```
choffmann, line 127: I think the name `treeClusterID` can be misleading as it holds an actual tree cluster entity. I don't want to be the guy who complains about variable names, but the name is not well chosen

internal/service/domain/tree/tree.go
```diff
@@ -121,32 +124,43 @@ func (s *TreeService) Create(ctx context.Context, treeCreate *entities.TreeCreat
...
-		treeClusterID, err := s.treeClusterRepo.GetByID(ctx, *treeCreate.TreeClusterID)
+		var err error
+		treeClusterID, err = s.treeClusterRepo.GetByID(ctx, *treeCreate.TreeClusterID)
 		if err != nil {
 			return nil, handleError(err)
 		}
-		fn = append(fn, tree.WithTreeCluster(treeClusterID))
 	}
 
+	var sensorID *entities.Sensor
```
choffmann, line 136: Same here. I think the name `sensorID` can be misleading

internal/service/domain/tree/tree.go
```diff
@@ -121,32 +124,43 @@ func (s *TreeService) Create(ctx context.Context, treeCreate *entities.TreeCreat
...
+		tree.Latitude = treeCreate.Latitude
+		tree.Longitude = treeCreate.Longitude
+
+		// Apply TreeCluster and Sensor if available
+		if treeClusterID != nil {
+			tree.TreeCluster = treeClusterID
+		}
+		if sensorID != nil {
+			tree.Sensor = sensorID
+		}
```
choffmann, line 159: You can do all the pre-defining of the tree cluster and sensor within this function. Then you can assign them directly to the `tree` entity you get in this function and you can also remove the misleading names `treeClusterID` and `sensorID`

internal/service/domain/tree/tree.go
```diff
@@ -178,42 +192,34 @@ func (s *TreeService) Update(ctx context.Context, id int32, tu *entities.TreeUpd
...
-		tree.WithDescription(tu.Description))
+	updatedTree, err := s.treeRepo.Update(ctx, id, func(tree *entities.Tree) (bool, error) {
+		tree.PlantingYear = tu.PlantingYear
+		tree.Species = tu.Species
+		tree.Number = tu.Number
+		tree.Latitude = tu.Latitude
+		tree.Longitude = tu.Longitude
+		tree.Description = tu.Description
+		tree.TreeCluster = treeCluster
+		tree.Sensor = sensor
```
choffmann, line 219: Same here, you can run the pre-definitions within this function and assign them directly to the tree entity

internal/storage/postgres/tree/create.go
```diff
@@ -25,48 +27,83 @@ func defaultTree() entities.Tree {
...
-	}
+func (r *TreeRepository) Create(ctx context.Context, createFn func(*entities.Tree) (bool, error)) (*entities.Tree, error) {
+	return r.createWithTx(ctx, createFn, nil)
+}
 
-	if err := r.validateTreeEntity(&entity); err != nil {
-		return nil, err
-	}
+func (r *TreeRepository) CreateAndLinkImages(ctx context.Context, createFn func(*entities.Tree) (bool, error)) (*entities.Tree, error) {
+	return r.createWithTx(ctx, createFn, func(ctx context.Context, tree *entities.Tree) error {
```
choffmann, line 35: Do we need this separate function? I would put this function together and if `tree.Image != nil` just assume that an image should be inserted

internal/storage/postgres/tree/update.go
```diff
@@ -3,51 +3,89 @@ package tree
...
-	}
+func (r *TreeRepository) Update(ctx context.Context, id int32, updateFn func(*entities.Tree) (bool, error)) (*entities.Tree, error) {
+	return r.updateWithTx(ctx, id, updateFn, nil)
+}
 
-	for _, fn := range tFn {
-		fn(entity)
-	}
+func (r *TreeRepository) UpdateWithImages(ctx context.Context, id int32, updateFn func(*entities.Tree) (bool, error)) (*entities.Tree, error) {
+	return r.updateWithTx(ctx, id, updateFn, func(ctx context.Context, tree *entities.Tree, updatedEntity *entities.Tree) error {
```
choffmann, line 18: Same here as in the create function

choffmann (changes requested):

choffmann:
Oh and please remove the now unused functions in `storage/postgres/tree/tree.go` 

`WithSensor(), WithSpecies(), ...`

internal/storage/postgres/tree/create.go
Mohammad96Assaf: This function has been created since the beginning of the project. Maybe its purpose is to be used somewhere else, not just like the `Create` function?

If we decide to remove it, we would also need to delete it from the interface and adjust the tree repository tests accordingly.

Are you sure that we don't use/call it separately? Should I go ahead and remove it?

internal/storage/postgres/tree/update.go
Mohammad96Assaf: The same answer as above.

internal/service/domain/sensor/sensor.go
```diff
@@ -166,7 +165,10 @@ func (s *SensorService) MapSensorToTree(ctx context.Context, sen *entities.Senso
 	}
 
 	if nearestTree != nil {
-		_, err = s.treeRepo.Update(ctx, nearestTree.ID, tree.WithSensor(sen))
+		_, err = s.treeRepo.Update(ctx, nearestTree.ID, func(tree *entities.Tree) (bool, error) {
+			tree.Sensor = sen
+			return true, nil
+		})
```
choffmann, line 171: Can you add a debug log that the sensor has been updated, something like this:
```suggestion
		_, err = s.treeRepo.Update(ctx, nearestTree.ID, func(tree *entities.Tree) (bool, error) {
			tree.Sensor = sen
			log.Debug("update sensor on tree", "tree_id", tree.ID, "sensor_id", sen.ID)
			return true, nil
		})
```

internal/service/domain/tree/handle_new_sensor_data.go
```diff
@@ -25,8 +24,11 @@ func (s *TreeService) HandleNewSensorData(ctx context.Context, event *entities.E
 		log.Debug("sensor status has not changed", "sensor_status", status)
 		return nil
 	}
+	newTree, err := s.treeRepo.Update(ctx, t.ID, func(s *entities.Tree) (bool, error) {
+		s.WateringStatus = status
+		return true, nil
+	})
```
choffmann, line 30: Please add a debug log here as well

internal/storage/postgres/tree/create.go
choffmann: Yes I know this function exits since the beginning of the project but a lot has changed since then and the way the create and update functions are set up, I don't see any added value from this function.

choffmann (changes requested):

internal/service/domain/tree/handle_new_sensor_data.go
```diff
@@ -25,8 +24,12 @@ func (s *TreeService) HandleNewSensorData(ctx context.Context, event *entities.E
 		log.Debug("sensor status has not changed", "sensor_status", status)
 		return nil
 	}
+	newTree, err := s.treeRepo.Update(ctx, t.ID, func(s *entities.Tree) (bool, error) {
+		slog.Debug("updating tree watering status", "prev_status", t.WateringStatus, "new_status", status)
```
choffmann, line 28:
```suggestion
		log.Debug("updating tree watering status", "prev_status", t.WateringStatus, "new_status", status)
```

choffmann (changes requested):

internal/service/domain/tree/handle_new_sensor_data.go
```diff
@@ -25,15 +23,17 @@ func (s *TreeService) HandleNewSensorData(ctx context.Context, event *entities.E
...
+		return true, nil
+	})
 
-	newTree, err := s.treeRepo.Update(ctx, t.ID, tree.WithWateringStatus(status))
 	if err != nil {
 		log.Error("failed to update tree with new watering status", "tree_id", t.ID, "watering_status", status, "err", err)
 		return err
 	}
 
-	slog.Info("updating tree watering status", "prev_status", t.WateringStatus, "new_status", status)
```
choffmann, line 35: Why did you remove this info log?

Mohammad96Assaf: I think it's enough to add it here?

```
	newTree, err := s.treeRepo.Update(ctx, t.ID, func(s *entities.Tree) (bool, error) {
		log.Debug("updating tree watering status", "prev_status", t.WateringStatus, "new_status", status)
		s.WateringStatus = status
		return true, nil
	})
```

choffmann: Yes I know you have moved inside the transaction but its a debug log. I would keep the debug log there, maybe with a message that it is inside the transaction and a separate info log when the update is successful

internal/storage/storage.go
```diff
@@ -165,13 +165,11 @@ type TreeRepository interface {
...
 	GetBySensorIDs(ctx context.Context, ids ...string) ([]*entities.Tree, error)
 
-	UpdateWithImages(ctx context.Context, id int32, fFn ...entities.EntityFunc[entities.Tree]) (*entities.Tree, error)
 	DeleteAndUnlinkImages(ctx context.Context, id int32) error
 	UnlinkAllImages(ctx context.Context, id int32) error
 	UnlinkTreeClusterID(ctx context.Context, treeClusterID int32) error
 	UnlinkSensorID(ctx context.Context, sensorID string) error
-	UnlinkImage(ctx context.Context, treeID, imageID int32) error
-	CreateAndLinkImages(ctx context.Context, tcFn ...entities.EntityFunc[entities.Tree]) (*entities.Tree, error)
+	UnlinkImage(ctx context.Context, flowerbedID, imageID int32) error
```
choffmann, line 172: Why is the flowerbed id here? @doriengr has removed all relevant code for the flowerbed in #407

doriengr: The flowerbedID should be renamed to treeID. I think it was not rebased correctly

Mohammad96Assaf: Oh, it was a mistake during rebasing.

choffmann (approved):
:+1:

PR #384, opened 2025-01-23
feat: improve log messages

### LINKED ISSUE #90: Improve error messages
We do not have good error messages yet. Improve error messages. I've tried to build a unified message system in `store.go`, but it's outdated. 

We also need to improve logging.

choffmann:
- add new slog handler `console`
- add debug, info, warn and error logging in application
- improve map errors in storage and service layer

image

To enable console logger, make sure to set `server.logs.format` to `console` in the config file

Close #90

https://github.com/green-ecolution/green-ecolution-backend/wiki/Logging-Styleguide

### DIFF
--- a/internal/server/http/handler/v1/region/handler_test.go
+++ b/internal/server/http/handler/v1/region/handler_test.go
@@ -10,8 +10,8 @@ import (
 	"github.com/green-ecolution/green-ecolution-backend/internal/entities"
 	serverEntities "github.com/green-ecolution/green-ecolution-backend/internal/server/http/entities"
 	"github.com/green-ecolution/green-ecolution-backend/internal/server/http/handler/v1/region"
+	"github.com/green-ecolution/green-ecolution-backend/internal/service"
 	serviceMock "github.com/green-ecolution/green-ecolution-backend/internal/service/_mock"
-	"github.com/green-ecolution/green-ecolution-backend/internal/storage"
 	"github.com/green-ecolution/green-ecolution-backend/internal/utils"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/mock"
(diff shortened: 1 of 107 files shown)


choffmann:
I've commented out the tests that check the specific error message because I don't think we're ready to implement that in our tests yet. We are still developing this application and the error messages are still not good

internal/server/http/handler/v1/region/handler_test.go
```diff
@@ -167,7 +167,7 @@ func TestGetRegionByID(t *testing.T) {
 		mockRegionService.EXPECT().GetByID(
 			mock.Anything,
 			int32(1),
-		).Return(nil, storage.ErrRegionNotFound)
+		).Return(nil, service.NewError(service.NotFound, "region not found"))
```
doriengr, line 170: Why is it everywhere else "not found" and here a specific error?

doriengr:
> I've commented out the tests that check the specific error message because I don't think we're ready to implement that in our tests yet. We are still developing this application and the error messages are still not good

I have a different opinion on this. If the error messages are going to change then we have to update the tests again. Just not checking the errors is not something I like to see.

doriengr (approved):

internal/server/http/handler/v1/region/handler_test.go
choffmann: This is a test. It doesn't matter what is defined here

choffmann:
> > I've commented out the tests that check the specific error message because I don't think we're ready to implement that in our tests yet. We are still developing this application and the error messages are still not good
> 
> I have a different opinion on this. If the error messages are going to change then we have to update the tests again. Just not checking the errors is not something I like to see. We can merge it as it is, but let's not finish the project with this state 🙏
> 
> Can we also create an issue for this so we do not forget?

I think we have too many fine-grained tests for our current status. Yes, everything should be in place by the end, but we are making our lives much harder than necessary.

internal/service/domain/sensor/mqtt.go
```diff
@@ -41,9 +46,19 @@ func (s *SensorService) HandleMessage(ctx context.Context, payload *domain.MqttP
 	return sensorData, nil
 }
 
+// TODO: missleading name, because its update or create the sensor
```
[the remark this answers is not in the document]
choffmann: I fixed it right away

PR #388, opened 2025-01-31
Feature/add watering plan pagination

### LINKED ISSUE #374: Implement pagination to watering plan handler

doriengr:
Close #374

### DIFF
--- a/internal/server/http/handler/v1/watering_plan/handler.go
+++ b/internal/server/http/handler/v1/watering_plan/handler.go
@@ -20,22 +20,22 @@ var (
 	wateringPlanMapper = generated.WateringPlanHTTPMapperImpl{}
 )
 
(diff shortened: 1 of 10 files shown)


internal/server/http/handler/v1/watering_plan/handler.go
```diff
@@ -33,12 +34,12 @@ var (
 // @Router			/v1/watering-plan [get]
 // @Param			page	query	string	false	"Page"
 // @Param			limit	query	string	false	"Limit"
```
choffmann, line 36: Page and limit are integers, right? We can specify types in Swagger
```suggestion
// @Param			page	query	int	false	"Page"
// @Param			limit	query	int	false	"Limit"
```

choffmann (approved):
Just had a quick look, didn't test it out...

PR #390, opened 2025-01-31
Feature/add vehicle pagination

### LINKED ISSUE #375: Implement pagination to vehicle handler

### LINKED ISSUE #397: Move logic in handler in GetAllVehicle to service
In the handler is a function `GetAllVehicles` that has to much logic in it which needs to move to the service

```go
if vehicleTypeStr != "" {
	vehicleType := domain.ParseVehicleType(vehicleTypeStr)
	if vehicleType == domain.VehicleTypeUnknown {
		return errorhandler.HandleError(service.NewError(service.BadRequest, "invalid vehicle type"))
	}
	domainData, err = svc.GetAllByType(ctx, vehicleType)
} else {
	domainData, err = svc.GetAll(ctx, provider)
} // TODO: move to service

```

doriengr:
Close #375 
Close #397

### DIFF
--- a/internal/server/http/handler/v1/vehicle/handler.go
+++ b/internal/server/http/handler/v1/vehicle/handler.go
@@ -17,23 +17,23 @@ var (
 	vehicleMapper = generated.VehicleHTTPMapperImpl{}
 )
 
-//	@Summary		Get all vehicles
-//	@Description	Get all vehicles
-//	@Id				get-all-vehicles
-//	@Tags			Vehicle
-//	@Produce		json
-//	@Success		200	{object}	entities.VehicleListResponse
-//	@Failure		400	{object}	HTTPError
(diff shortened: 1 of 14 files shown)


internal/server/http/handler/v1/vehicle/handler.go
```diff
@@ -30,13 +31,14 @@ var (
 // @Router			/v1/vehicle [get]
 // @Param			page	query	string	false	"Page"
 // @Param			limit	query	string	false	"Limit"
```
choffmann, line 33: Page and limit are integers, right? We can specify types in Swagger 

```suggestion
// @Param			page	query	int	false	"Page"
// @Param			limit	query	int	false	"Limit"
```

choffmann (approved):

choffmann:
Just had a quick look, didn't test it out...

PR #395, opened 2025-02-05
feat: add provider to entities

### LINKED ISSUE #369: Implement database connection for tree register Flensburg

choffmann:
add provider and additional info fields to `tree`, `treecluster`, `sensor`, `vehicle` and `watering plan`. These fields allow an external provider to create an entity in our system and store specific metadata about that entity in an additional info field

Ref #369


internal/server/http/handler/v1/sensor/handler.go
```diff
@@ -119,6 +121,8 @@ func DeleteSensor(svc service.SensorService) fiber.Handler {
 	}
 }
 
+// TODO: Create / Update Sensor
```
doriengr, line 124: Do we need those handlers?

internal/server/http/handler/v1/vehicle/handler.go
```diff
@@ -28,10 +28,11 @@ var (
 // @Failure		404	{object}	HTTPError
 // @Failure		500	{object}	HTTPError
 // @Router			/v1/vehicle [get]
-// @Param			page	query	string	false	"Page"
-// @Param			limit	query	string	false	"Limit"
-// @Param			status	query	string	false	"Status"
-// @Param			type	query	string	false	"Vehicle Type"
+// @Param			page		query	string	false	"Page"
+// @Param			limit		query	string	false	"Limit"
+// @Param			status		query	string	false	"Status"
```
doriengr, line 33: We do not need a status as a parameter here

internal/server/http/handler/v1/vehicle/handler.go
```diff
@@ -40,15 +41,16 @@ func GetAllVehicles(svc service.VehicleService) fiber.Handler {
...
 		if vehicleTypeStr != "" {
 			vehicleType := domain.ParseVehicleType(vehicleTypeStr)
 			if vehicleType == domain.VehicleTypeUnknown {
 				return errorhandler.HandleError(service.NewError(service.BadRequest, "invalid vehicle type"))
 			}
 			domainData, err = svc.GetAllByType(ctx, vehicleType)
 		} else {
-			domainData, err = svc.GetAll(ctx)
-		}
+			domainData, err = svc.GetAll(ctx, provider)
```
doriengr, line 52: What do we have to move to the server? Depending which function to call? Maybe create an issue for the todos? :)

internal/server/http/handler/v1/watering_plan/handler.go
```diff
@@ -31,14 +31,16 @@ var (
 // @Failure		404	{object}	HTTPError
 // @Failure		500	{object}	HTTPError
 // @Router			/v1/watering-plan [get]
-// @Param			page	query	string	false	"Page"
-// @Param			limit	query	string	false	"Limit"
-// @Param			status	query	string	false	"Status"
+// @Param			page		query	string	false	"Page"
+// @Param			limit		query	string	false	"Limit"
+// @Param			status		query	string	false	"Status"
```
doriengr, line 36: Same here: We do not need a status. I think its a copy paste mistake.

doriengr (commented):
I've only skimmed the PR, but it looks okay. Did I understand correctly that you added the additional info and the provider to be able to say that certain trees come from the DB, for example, and other trees are imported etc?

doriengr (approved):

internal/server/http/handler/v1/sensor/handler.go
choffmann: Yes, not for the frontend or so but have a RESTful backend api and the ability to register plugins, maybe there will be a plugin that would like to create or update sensor in our system

internal/server/http/handler/v1/vehicle/handler.go
choffmann: Yes, exactly

choffmann: #397

internal/server/http/handler/v1/sensor/handler.go
choffmann: #398

choffmann:
> I've only skimmed the PR, but it looks okay. Did I understand correctly that you added the additional info and the provider to be able to say that certain trees come from the DB, for example, and other trees are imported etc?

Yes, I needed a way to do an initial sync between the plugin and our system with only one source of truth

PR #396, opened 2025-02-07
Feature/Implement filter function in tree cluster

### LINKED ISSUE #83: Implement filter function

Mohammad96Assaf:
close #83


internal/entities/treecluster.go
```diff
@@ -45,3 +45,8 @@ type TreeClusterUpdate struct {
 	TreeIDs       []*int32
 	Name          string `validate:"required"`
 }
+
+type TreeClusterFilter struct {
```
doriengr, line 49: Is this best practice? I do not know myself …

doriengr:
Something is not working

image

What I expect:
1.  If I do not have pagination values and no filter activated, the endpoint should return everything
2. If I do have set pagination values but no filters activated, the endpoint should return every tree cluster but paginated
3. If I do not have pagination values but some filters activated, the endpoint should return everything but filtered
4. If I do have pagination values and filters are activated, the endpoint should return filtered tree cluster but the tree cluster are paginated

internal/server/http/handler/v1/treecluster/handler_test.go
```diff
@@ -177,6 +177,125 @@ func TestGetAllTreeCluster(t *testing.T) {
...
+		assert.Equal(t, http.StatusOK, resp.StatusCode)
+
+		var response serverEntities.TreeClusterListResponse
+		err = utils.ParseJSONResponse(resp, &response)
+		assert.NoError(t, err)
+
+		assert.Equal(t, len(expectedFiltered), len(response.Data))
+		for _, cluster := range response.Data {
+			assert.NotNil(t, cluster.Region)
+			assert.Equal(t, "Mürwik", cluster.Region.Name)
```
doriengr, line 253: Pagination values are not checked

internal/server/http/handler/v1/treecluster/handler_test.go
```diff
@@ -177,6 +177,125 @@ func TestGetAllTreeCluster(t *testing.T) {
...
+
+		var response serverEntities.TreeClusterListResponse
+		err = utils.ParseJSONResponse(resp, &response)
+		assert.NoError(t, err)
+
+		assert.Equal(t, len(expectedFiltered), len(response.Data))
+		for _, cluster := range response.Data {
+			assert.EqualValues(t, entities.WateringStatusModerate, cluster.WateringStatus)
+			assert.NotNil(t, cluster.Region)
+			assert.Equal(t, "Mürwik", cluster.Region.Name)
```
doriengr, line 294: Pagination values are not checked

internal/server/http/handler/v1/treecluster/handler.go
```diff
@@ -29,12 +33,29 @@ var (
...
+// @Param			region	query	string	false	"region name"
 // @Security		Keycloak
 func GetAllTreeClusters(svc service.TreeClusterService) fiber.Handler {
 	return func(c *fiber.Ctx) error {
 		ctx := c.Context()
 
-		domainData, totalCount, err := svc.GetAll(ctx)
+		statusQu := c.Query("status")
+		var wateringStatus domain.WateringStatus
+		if statusQu != "" {
```
doriengr, line 45: It should also be possible to filter by multiple values:

For example:
image

This filter for example should return all tree cluster with the watering status `good` OR `bad`, but the tree cluster do have to be placed in the region `Mürwik` OR `Fuerlund`.

But the join between the different filter options should be AND.

internal/storage/postgres/treecluster/get_test.go
```diff
@@ -143,11 +144,93 @@ func TestTreeClusterRepository_GetAll(t *testing.T) {
...
+
+		ctx := context.WithValue(context.Background(), "page", int32(1))
+		ctx = context.WithValue(ctx, "limit", int32(-1))
+
+		filter := entities.TreeClusterFilter{
+			WateringStatus: entities.WateringStatusGood,
+		}
+
+		// when
+		got, _, err := r.GetAll(ctx, filter)
```
doriengr, line 167: Please also check the `totalCount`

internal/storage/postgres/treecluster/get_test.go
```diff
@@ -143,11 +144,93 @@ func TestTreeClusterRepository_GetAll(t *testing.T) {
...
+
+		ctx := context.WithValue(context.Background(), "page", int32(1))
+		ctx = context.WithValue(ctx, "limit", int32(-1))
+
+		filter := entities.TreeClusterFilter{
+			Region: "Mürwik",
+		}
+
+		// when
+		got, _, err := r.GetAll(ctx, filter)
```
doriengr, line 193: Please also check the `totalCount`

internal/storage/postgres/treecluster/get_test.go
```diff
@@ -143,11 +144,93 @@ func TestTreeClusterRepository_GetAll(t *testing.T) {
...
+		ctx := context.WithValue(context.Background(), "page", int32(1))
+		ctx = context.WithValue(ctx, "limit", int32(-1))
+
+		filter := entities.TreeClusterFilter{
+			WateringStatus: entities.WateringStatusModerate,
+			Region:         "Mürwik",
+		}
+
+		// when
+		got, _, err := r.GetAll(ctx, filter)
```
doriengr, line 221: Please also check the `totalCount`

internal/storage/postgres/treecluster/get_test.go
```diff
@@ -143,11 +144,93 @@ func TestTreeClusterRepository_GetAll(t *testing.T) {
...
+		}
+	})
+
+	t.Run("should return tree clusters filtered by both watering status and region", func(t *testing.T) {
+		// given
+		suite.ResetDB(t)
+		suite.InsertSeed(t, "internal/storage/postgres/seed/test/treecluster")
+		r := NewTreeClusterRepository(suite.Store, mappers)
+
+		ctx := context.WithValue(context.Background(), "page", int32(1))
```
doriengr, line 212: There is a test case missing with paginated values

doriengr (changes requested):

internal/entities/treecluster.go
Mohammad96Assaf: Please try to write a complete comment and explain why you don’t see it as good practice. Do you have other suggestions?

doriengr: I am asking if this is best practise and if you did some research on it? I am wondering because I have never seen such an Implementation.

Mohammad96Assaf: We have multiple filter criteria, and moving them between layers as a struct is better than keeping each attribute as a separate parameter. Actually, I don't have a better idea. If you have a better plan, please let me know.

Mohammad96Assaf:
@doriengr ,I'm not sure why you're returning `totalCount` from the GetAll function in the storage layer. You already return a data list as the first return value, and if you need the total count later, you can just use `len(data)`. I don't think we need to return totalCount from `GetAll` in the storage layer.

Take a look at this test case:
```
assert.Empty(t, got)
assert.Equal(t, totalCount, int64(0))
```
Here, we’re checking if the list `got` is empty and also that the `total count` is 0, which demonstrates the redundancy.

doriengr:
> @doriengr ,I'm not sure why you're returning `totalCount` from the GetAll function in the storage layer. You already return a data list as the first return value, and if you need the total count later, you can just use `len(data)`. I don't think we need to return totalCount from `GetAll` in the storage layer.
> 
> Take a look at this test case:
> 
> ```
> assert.Empty(t, got)
> assert.Equal(t, totalCount, int64(0))
> ```
> 
> Here, we’re checking if the list `got` is empty and also that the `total count` is 0, which demonstrates the redundancy.

Please take a look at how paginations work. Total count is the count of all items in a table and the list of items are paginated. They are not the same as you can see in tests where pagination values are set. 

We need both values:)

internal/entities/treecluster.go
choffmann: What about storing the values in the request context like the pagination values?

Mohammad96Assaf: After looking into it, I found that it's not a good idea to store data in `context.Context`.
Context is mainly used for managing request `timeouts` and `cancellations`, not for storing arbitrary data.

Here are some sources that explain this better:

https://pauldigian.com/golang-and-context-an-explanation#contextcontext-to-store-values
https://www.calhoun.io/pitfalls-of-context-values-and-how-to-avoid-or-mitigate-them/

Instead of using context, we can use : 

-  a struct (like `TreeClusterFilter`) to group data, and I understood is a typical way.

- or `functional options (opts ...Option)`, but this requires creating a `WithFunc` for each value, and in the end, we still need to use a struct to group the data.


I would say we continue using a struct as a parameter, it's flexible and easy to use.

Mohammad96Assaf: What about the provider string that is used everywhere in the GetAll function?

`GetAll(ctx context.Context, provider string) ([]*T, int64, error)`

Here, we are using the same approach. I think placing this parameter inside a struct would be a cleaner solution.

We can create a generic filter struct for the generic GetAll function:
```

type Filter[T any] struct {
    Conditions T
    Provider   string 
}

type TreeClusterFilter struct {
    WateringStatus []WateringStatus
    Region         []string
}

type TreeClusterService interface {
    GetAll(ctx context.Context, filter entities.Filter[entities.TreeClusterFilter]) ([]*entities.TreeCluster, int64, error)
}
```
Example Usage:
```
filter := filters.Filter[entities.TreeClusterFilter]{
    Conditions: entities.TreeClusterFilter{
        WateringStatus: []string{"moderate", "good"},
        Region:         []string{"Mürwik", "Altstadt"},
    },
    Page:     1,
    Limit:    10,
    Provider: "",
}

result, count, err := treeClusterService.GetAll(ctx, filter)
```
what do you think about this ?

internal/storage/postgres/treecluster/get_test.go
```diff
@@ -165,11 +169,207 @@ func TestTreeClusterRepository_GetAll(t *testing.T) {
...
+		filter := entities.TreeClusterFilter{
+			WateringStatus: wateringstatues,
+			Region:         regionNames,
+			Provider:       "",
+		}
+
+		// when
+		got, totalCount, err := r.GetAll(ctx, filter)
+
+		fmt.Println(got)
```
doriengr, line 329: Here are some leftover print lines

go.sum
choffmann, line 1: Please don't delete this file

internal/entities/treecluster.go
choffmann: Yes, we can move the provider attribute into the filter struct, but I wouldn't create a generic struct. You can create a filter struct that is implemented by every other specific filter struct

```go
type Filter struct {
    Provider   string 
}

type TreeClusterFilter struct {
    Filter
    WateringStatus []WateringStatus
    Region         []string
}

filter := TreeClusterFilter{
    WateringStatus: []string{"moderate", "good"},
    Region:         []string{"Mürwik", "Altstadt"},
    Filter: Filter{
        Provider: "",
    }
}
```

Mohammad96Assaf: I prefer to implement it in a separate issue.

xk0niR: Please use this: https://docs.gofiber.io/api/ctx/#queryparser

xk0niR:
Please remove go.mod and go.sum changes. Thanks.

choffmann:
@Mohammad96Assaf the `go.mod` and `go.sum` files are still not reseted and in this PR. Please reset the files and not regenerate then

internal/entities/treecluster.go
```diff
@@ -51,3 +51,9 @@ type TreeClusterUpdate struct {
 	Provider       string
 	AdditionalInfo map[string]interface{}
 }
+
+type TreeClusterQuery struct {
+	WateringStatus []WateringStatus `query:"status"`
+	Region         []string         `query:"region"`
+	Query          `query:""`
```
choffmann, line 58: I don't think you need the annotation and the end in this line
```suggestion
	Query
```

internal/entities/vehicle.go
```diff
@@ -72,6 +72,11 @@ type VehicleUpdate struct {
 	AdditionalInfo map[string]interface{}
 }
 
+type VehicleQuery struct {
+	Type  string `query:"type"`
+	Query `query:""`
```
choffmann, line 77:
```suggestion
	Query
```

choffmann (changes requested):
A wrong user request (here for example status=foo) should return status code 400 Bad Request

image

Mohammad96Assaf:
@doriengr @choffmann , the PR is ready for review again. To avoid redundant code, I need this PR to proceed with #84.

doriengr (approved):
Endpoints that I tested:

1. Tree Cluster (with pagination and filters)
2. Sensor (with pagination)
3. Tree (with pagination)
4. Vehicle (with pagination and filters)
5. Watering Plans (with pagination)

choffmann (approved):
I'll trust @doriengr

PR #401, opened 2025-02-08
feat: add expiry field in client token

choffmann:
add expiry field in client token and improve generate client packages


doriengr (approved):
🔥

PR #402, opened 2025-02-08
feat: add unregister plugin

choffmann:


internal/server/http/handler/v1/plugin/handler.go
```diff
@@ -99,6 +99,28 @@ func registerPlugin(svc service.PluginService) fiber.Handler {
...
+// @Tags			Plugin
+// @Produce		json
+// @Success		204
+// @Failure		401	{object}	HTTPError
+// @Failure		404	{object}	HTTPError
+// @Failure		500	{object}	HTTPError
+// @Router			/v1/plugin/{plugin_slug}/unregister [post]
+// @Param			plugin_slug	path	string	true	"Slug of the plugin"
+// @Security		Keycloak
+func unregisterPlugin(svc service.PluginService) fiber.Handler {
```
doriengr, line 114: Why is this not `UnregisterPlugin`?

pkg/plugin/http_func.go
```diff
@@ -95,6 +95,58 @@ func (w *PluginWorker) Register(ctx context.Context, clientID, clientSecret stri
...
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusNoContent {
+		return errors.New("failed to unregister plugin")
+	}
+
+	return nil
+}
+
+// TODO: implement
```
doriengr, line 145: You know me: that sounds like a good new issue ;)

doriengr (approved):
LGTM

pkg/plugin/http_func.go
choffmann: Shit, I didn't really want to push this... Oh well :face_with_peeking_eye:

internal/server/http/handler/v1/plugin/handler.go
choffmann: Yes, it can also be capitalized, but actually it would be correct to write it in lower case, also the other methods because they are only used in the route which is in the same package

pkg/plugin/http_func.go
choffmann: #404

PR #405, opened 2025-02-09
Fix/mapdata config

Zoka-tech:
I am not sure how to fix the merge conflicts.
The storage file is easy of course. But I am not sure how to fix the config. I added the mapdata to the config, so I need this file. Do I just take my version and discard the rest?


Zoka-tech:
I am not sure how to fix the merge conflicts.
The storage file is easy of course. But I am not sure how to fix the config. I added the mapdata to the config, so I need this file. Do just take my version and discard the rest?

config/config.dev.enc.yaml
```diff
@@ -1,66 +1,76 @@
...
+map:
+    center:
+        - ENC[AES256_GCM,data:OdVy0R4CWPCFKFNJojpPUT4J,iv:pD5UlCN8AqzFQwV4p7e/ZuLPNTQQkKXEqPuoWXeYwvw=,tag:pfzK0d1l0nPnSJFJ9KQ2Hw==,type:float]
+        - ENC[AES256_GCM,data:FZ7ylEiIAnwqaTErDrBduw==,iv:9DgaUtV5Q/I8PL2frWUAYX5PaXz5aJm76deUUDx6GUs=,tag:cZ9Cfj38r68vSyLTfq9seg==,type:float]
+    bound_south_west:
+        - ENC[AES256_GCM,data:RZdkgHI=,iv:Jt+aNxVgjye0xdFmv/9aUvz3vgk0cGEmNGHO9HwBKUo=,tag:FyB463IfojTUpHkyZxb98A==,type:float]
+        - ENC[AES256_GCM,data:9mSL,iv:i/0ou+KWUuZNE5HzJ/8b+YHeRrZwit24Nxi84Zdjs8A=,tag:0fZiO9k5GD6Yn4G8bVZz5g==,type:float]
+    bound_north_east:
+        - ENC[AES256_GCM,data:fgsfuoA=,iv:Ls4AbZqa+lMWWBHZXzOnD+2FhHuBS43jkS6wcdEAeUU=,tag:/0upukfGhxAgKF5Y5p1ePA==,type:float]
+        - ENC[AES256_GCM,data:ybH3,iv:EmXoewDQQ2cWr33a80f7Vw6MROQAUwqxGuzFH2XtUKw=,tag:OUMb1BfbC3L4f73m+bbpwQ==,type:float]
```
choffmann, line 73: Bounding boxes are standardised and normally an array of 4 elements. So instead of specific specifications like `bound_south_west` or `bound_north_east` I would consider something like this: 

```yaml
map:
    center: [54.792277136221905, 9.43580607453268]
    bbox: [54.85, 9.6, 54.73, 9.3]
```

image
https://wiki.openstreetmap.org/wiki/Bounding_box

.docker/docker-compose.infra.yaml
choffmann, line 1: Please do not commit these changes to git, this attribute will be overridden on `make infra/up` if your user id and group id are not `1000:1000` by default. I want to keep `1000:1000` as default

internal/config/app.go
```diff
@@ -59,6 +59,12 @@ type RoutingVroomConfig struct {
 	Host string `mapstructure:"host"`
 }
 
+type MapConfig struct {
+	Center         []float64 `mapstructure:"center"`
+	BoundSouthWest []float64 `mapstructure:"bound_south_west"`
+	BoundNorthEast []float64 `mapstructure:"bound_north_east"`
```
choffmann, line 65: Same here, I would prefer an simple array
```suggestion
	BoundBox []float64 `mapstructure:"bbox"`
```

internal/entities/info.go
```diff
@@ -30,3 +31,9 @@ type Server struct {
 	Interface string
 	Uptime    time.Duration
 }
+
+type Map struct {
+	Center         GeoJSONLocation
+	BoundSouthWest GeoJSONLocation
+	BoundNorthEast GeoJSONLocation
```
choffmann, line 38: These attributes are not GeoJSON, GeoJSON is different. This should also be an array as well

GeoJson definition: https://datatracker.ietf.org/doc/html/rfc7946

.docker/docker-compose.infra.yaml
choffmann: I will reset this for you in the Git history :+1:

choffmann (changes requested):
Looks good so far :trophy: I will take care of the merge conflicts and git problems for you :+1:

internal/entities/info.go
choffmann: Ahh okay, you are using the latitude and longitude fields of this struct, I think a simple array will do the job here.

choffmann:
@Zoka-tech I have rebased your branch onto develop and added the map values to the stage and prod config as well. When you edit this PR, please remember to update the values in the stage and prod configs as well.

.docker/docker-compose.infra.yaml
choffmann, line 1: Do not delete, please reset...

choffmann (changes requested):
Dev config is now in the correct format, but prod and stage are not

xk0niR:
Wait until https://github.com/green-ecolution/green-ecolution-frontend/pull/377 is merged.

choffmann (approved):
@Zoka-tech I just had to pass the config in the default docker config files to make the deployments work. I also changed the error message and use the officle bbox value for flensburg.

EDIT: And use a fixed size array for center and bbox instead of a slice, because these values always have the same size of 2 for center and 4 for bbox.

PR #406, opened 2025-02-09
Feature/add pagination to trees

doriengr:
Close https://github.com/green-ecolution/green-ecolution-backend/issues/371


internal/server/http/handler/v1/tree/handler.go
```diff
@@ -36,8 +37,7 @@ var (
 func GetAllTrees(svc service.TreeService) fiber.Handler {
 	return func(c *fiber.Ctx) error {
 		ctx := c.Context()
-		provider := c.Query("provider")
-		domainData, err := svc.GetAll(ctx, provider)
+		domainData, totalCount, err := svc.GetAll(ctx, c.Query("provider"))
```
choffmann, line 40: Unfortunately, we have to clone the value if we want to use it outside the handler.... (which I myself have also forgotten)

```suggestion
		domainData, totalCount, err := svc.GetAll(ctx, strings.Replace(c.Query("provider")))
```

image

choffmann (changes requested):

choffmann (approved):
:+1:

PR #411, opened 2025-02-10
Feature/get all sensor data by sensor

doriengr:
Close https://github.com/green-ecolution/green-ecolution-backend/issues/412

### DIFF
--- a/internal/storage/postgres/sensor/get.go
+++ b/internal/storage/postgres/sensor/get.go
@@ -56,6 +56,23 @@ func (r *SensorRepository) GetAll(ctx context.Context, provider string) ([]*enti
 	return data, totalCount, nil
 }
 
+func (r *SensorRepository) GetAllDataByID(ctx context.Context, id string) ([]*entities.SensorData, error) {
+	log := logger.GetLogger(ctx)
+	rows, err := r.store.GetAllSensorDataByID(ctx, id)
+	if err != nil {
+		log.Debug("failed to get all sensor data by sensor id in db", "error", err, "sensor_id", id)
+		return nil, r.store.MapError(err, sqlc.Sensor{})
+	}
+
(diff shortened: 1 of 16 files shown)


doriengr:
> @doriengr please fix the migrations PR test.

It is already fixed in the Main. I only have to rebase. You can already make a review

internal/storage/postgres/sensor/get_test.go
```diff
@@ -167,6 +167,73 @@ func TestSensorRepository_GetAll(t *testing.T) {
 	})
 }
 
+func TestSensorRepository_GetAllDataById(t *testing.T) {
```
Mohammad96Assaf, line 170: Please ensure a tests case cover these scenarios: 

- `should return error when sensor not found`
- `should return error when sensor id is empty`

internal/storage/postgres/sensor/get.go
```diff
@@ -56,6 +56,23 @@ func (r *SensorRepository) GetAll(ctx context.Context, provider string) ([]*enti
 	return data, totalCount, nil
 }
 
+func (r *SensorRepository) GetAllDataByID(ctx context.Context, id string) ([]*entities.SensorData, error) {
```
Mohammad96Assaf, line 59: To cover the case where the function returns an error if the sensor for the given ID is not found, I think you should call `getById ` first and then return its data.

Mohammad96Assaf (changes requested):
image
Currently, when an invalid sensor ID is provided, the API returns` 200 OK` with an empty array. 
Please return `404 Not Found` instead of `200 OK` when the `sensor ID` does not exist.
And Please if possible, apply the same fix to `GetLatestSensorDataBySensorID`.

Mohammad96Assaf (approved):

PR #416, opened 2025-02-12
feat: change pagination definition in swagger

choffmann:
- Pagination can be null, so I added `omitempty` and `validate:"optional"` so that Swagger can handle it
- if there are no values set for `limit` and `page` in the context, default values are returned instead of error

### DIFF
--- a/internal/server/http/entities/user.go
+++ b/internal/server/http/entities/user.go
@@ -38,8 +38,7 @@ type UserResponse struct {
 } // @Name User
 
 type UserListResponse struct {
-	Data       []*UserResponse `json:"data"`
-	Pagination *Pagination     `json:"pagination,omitempty" validate:"optional"`
+	Data []*UserResponse `json:"data"`
 } // @Name UserList
 
 type UserRegisterRequest struct {
--- a/internal/server/http/entities/user.go
+++ b/internal/server/http/entities/user.go
@@ -39,7 +39,7 @@ type UserResponse struct {
 
 type UserListResponse struct {
 	Data       []*UserResponse `json:"data"`
-	Pagination Pagination      `json:"pagination"`
+	Pagination *Pagination     `json:"pagination,omitempty" validate:"optional"`
 } // @Name UserList
 
 type UserRegisterRequest struct {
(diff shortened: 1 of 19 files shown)


internal/utils/pagination/pagination.go
```diff
@@ -14,8 +14,8 @@ func GetValues(ctx context.Context) (page, limit int32, err error) {
 	limit, limitOK := ctx.Value("limit").(int32)
 
 	if !pageOk || !limitOK {
-		log.Debug("pagination values not found in context", "page", ctx.Value("page"), "limit", ctx.Value("limit"))
-		return 0, 0, storage.ErrPaginationValueInvalid
+		page = 1
```
doriengr, line 17: But when is the pagination then null, when you set default values? I wrote it that way because we can not have a pagination on the map for example ...

internal/server/http/entities/user.go
```diff
@@ -39,7 +39,7 @@ type UserResponse struct {
 
 type UserListResponse struct {
 	Data       []*UserResponse `json:"data"`
-	Pagination Pagination      `json:"pagination"`
+	Pagination *Pagination     `json:"pagination,omitempty" validate:"optional"`
 } // @Name UserList
```
doriengr, line 43: User can not have a pagination. This can be deleted.

doriengr (changes requested):

internal/utils/pagination/pagination.go
choffmann: These are the defaults you set in the middleware if the request doesn't have the query parameter set in the request. And when I call the endpoint with the values `page=1` and `limit=-1` I get all entities with pagination `null`

image

I implemented this because these functions will also not be called in the context of a request but in our application itself, so in this case I want to have all the values

image

internal/server/http/handler/v1/region/handler.go
```diff
@@ -39,7 +39,7 @@ func GetAllRegions(svc service.RegionService) fiber.Handler {
 
 		return c.JSON(entities.RegionListResponse{
 			Regions:    dto,
-			Pagination: entities.Pagination{}, // TODO: Handle pagination
+			Pagination: nil, // TODO: Handle pagination
```
doriengr, line 42: Do we need this pagination?

doriengr (approved):

internal/server/http/handler/v1/region/handler.go
choffmann: Yah, why not? :man_shrugging:

PR #417, opened 2025-02-14
Feature/scheduler for updating watering plan status

### LINKED ISSUE #291: Schedular to check if watering plans are not competed

doriengr:
Close #291


choffmann (approved):
:+1:

PR #422, opened 2025-02-19
fix: check if region is not nil before accessing it

### LINKED ISSUE #421: Empty tree cluster crashes backend
When trying to create a tree cluster without a tree the backend will crash.

EDIT: It doesn´t seem the lack of a tree is the problem but rather a missing location.
E.g. create a tree cluster with 1 tree which has no region (to test it, create a tree inside water).
Same error.

choffmann:
Close #421


choffmann:
I wanted to test it, but the logic is in the transaction. At the moment I don't have a good idea how to test it.

xk0niR (changes requested):
Semi-Fixed.

- New Tree cluster without tree: Ok
- New Tree cluster with one tree without region: Backend Crash

xk0niR (approved):
Everything works now!
:100:

PR #428, opened 2025-02-21
feat: remove readonly attribute from tree entity

### LINKED ISSUE #399: Remove readonly attribute in tree entity
The `readonly` attribute in the tree entity is obsolete because there is a provider field in the main entities `tree`, `treecluster`, `sensor`, `vehicle`, and `watering plan`. If the provider is not empty, we can assume that this object is read-only

choffmann:
Close #399


doriengr (approved):
Wuhuuuu

PR #433, opened 2025-02-23
feat: archive vehicles

### LINKED ISSUE #430: WateringPlan API returns »404 entitiy not found« and crashes frontend
How to reproduce:
1. Create a watering plan with a transporter
2. Delete the transporter
3. Now the whole watering plan frontend stuff depending on watering plans are crashing.

choffmann:
Close #430

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -9,7 +9,7 @@ linters-settings:
           - pkg: "github.com/sirupsen/logrus"
             desc: logging is allowed only by logutils.Log
   dupl:
-    threshold: 100
+    threshold: 150
   funlen:
     lines: -1 # the number of lines (code + empty lines) is not a right metric and leads to code without empty line or one-liner.
     statements: 50
--- a/internal/entities/vehicle.go
+++ b/internal/entities/vehicle.go
@@ -25,6 +25,7 @@ type Vehicle struct {
 	ID             int32
 	CreatedAt      time.Time
 	UpdatedAt      time.Time
+	ArchivedAt     time.Time
 	NumberPlate    string
 	Description    string
 	WaterCapacity  float64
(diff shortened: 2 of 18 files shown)


xk0niR (changes requested):
- Create a new vehicle: OK
grafik

- Create new watering plan: OK
grafik

- Delete vehicle: OK

- Open watering plan: ERROR

I am unable to open the watering plan. Backend shows
"error="entity not found: WateringPlan"

Also I created another vehicle and since then I am unable to open watering plan page
"DEBUG internal/storage/postgres/watering_plan/get.go:92 failed to get linked vehicle entity by id and vehicle type user_id="9fddda2f-657d-4b81-b852-8a333ea2b4a1" error="no rows in result set" watering_plan_id="9" vehicle_type="trailer""

Edit: Last message may not related to another vehicle being created. It could be watering plan overview was loading fine because broken watering plan was on 2nd page. If another watering plan gets deleted and the broken watering plan moves to 1st page you can´t load the site anymore.

choffmann:
@xk0niR I haven't touched the frontend yet. The delete endpoint is still active and will delete the entity from the database. To test it, you need to manually archive the vehicle via the backend api route `POST /api/v1/vehicle/archive/:id`

choffmann:
> @choffmann Since we alter database table doesn´t the seed also to be updated?

@xk0niR No, I have implemented an `archive_at` attribute in the db which is set to null by default

xk0niR (approved):
Tested using swagger with vehicle id 3
Archive:
grafik

List all archived vehicles:
grafik

Vehicle is also no longer inside get-all-vehicles and get-vehicle-by-id also works.
Works :1st_place_medal:

PR #436, opened 2025-02-24
bugfix: fix conn busy err

### LINKED ISSUE #427: Handle multiple connections on work load
When there is a lot of work (in this case a whole import from the tbz-baumkataster-plugin which creates many trees for some time) and a request goes to the trees, it returns `500 conn busy`

Image

### LINKED ISSUE #429: Finishing a watering plan does only update sometimes the watering status to »just watered«

choffmann:
Close #427 
Close #429 

This issue fixes common issues with multiple users working on the same object at the same time


xk0niR (approved):
What I tested:
Fill my database with tbz trees: OK

--- Easy Tests
vehicle: create/update/delete: OK
tree: create/update/delete: OK
treecluster: create/update/delete: OK
watering plan: create/update/delete: OK

sensors: There is no create or update (except manually editing database)
I deleted a sensor successfully with swagger: OK

--- Extended tests
Created treecluster t1 and treeclusgter t2

Tree from t1 to t2:
- gets removed from t1: OK
- gets added to t2: OK

I have also tested other functions without any error. :1st_place_medal:

PR #440, opened 2025-02-25
fix: adjust TTN attribute names

### LINKED ISSUE #439: Adjust the conversion of the MQTT payload to make it compatible with TTN data

Mohammad96Assaf:
close #439


choffmann:
Please also edit the dev, stage, prod and docker default config files and change the mqtt topic

choffmann (approved):
I implemented a safe cast and remove the condition when lat or long is zero the sensor will be inserted in the db now

PR #442, opened 2025-02-26
fix: update watering status of tc while creating it

### LINKED ISSUE #441: Tree Cluster doesn´t update with 1 tree
Create a tree cluster with one tree: The status will always be "Unbekannt" even if the tree is "In Ordnung".

Mohammad96Assaf:
close #441

### DIFF
--- a/internal/entities/events.go
+++ b/internal/entities/events.go
@@ -11,6 +11,7 @@ const (
 	EventTypeCreateTree         EventType = "create tree"
 	EventTypeDeleteTree         EventType = "delete tree"
 	EventTypeUpdateTreeCluster  EventType = "update tree cluster"
+	EventTypeCreateTreeCluster  EventType = "create tree cluster"
 	EventTypeNewSensorData      EventType = "receive sensor data"
 	EventTypeUpdateWateringPlan EventType = "update watering plan"
 )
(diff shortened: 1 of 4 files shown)


doriengr:
This is not working, right? You do not set the watering status anywhere …

Mohammad96Assaf:
> This is not working, right? You do not set the watering status anywhere …

I don't have to set the watering status. I only created a "create-event" for the "create tree cluster function".
It is working for #441

Mohammad96Assaf:
@choffmann I have already tested it with @doriengr and @xk0niR. It works locally for me

choffmann:
@Mohammad96Assaf So please tell me what I'm doing wrong

[Kooha-2025-03-16-18-51-57.webm](https://github.com/user-attachments/assets/b179f44a-1193-48db-b286-2e359f599fff)

xk0niR:
@choffmann this is strange... I used his branch + latest develop and it is working for me.

tcUpdate-opt

choffmann:
And another question, if I do the same step with the current develop branch, this will work, so what does this PR fix?

xk0niR:
> Okay, maybe this is a special case to the seed data, but maybe we should deal with this special case

Ah you are right, a tree without a sensor shouldn´t have any status (except unknown).
We should deal with it because it can cause confusion later.

xk0niR:
> And another question, if I do the same step with the OK tree and the current develop branch, this will work, so what does this PR fix?

I think it doesn´t fix anything but this issue was more like caused by seed data and not testing different trees.
If I am correct this introduces treecluster create event but I guess it doesn´t is needed as a new tree will always trigger update? @Mohammad96Assaf may has more information.

Mohammad96Assaf:
> > And another question, if I do the same step with the OK tree and the current develop branch, this will work, so what does this PR fix?
> 
> I think it doesn´t fix anything but this issue was more like caused by seed data and not testing different trees. If I am correct this introduces treecluster create event but I guess it doesn´t is needed as a new tree will always trigger update? @Mohammad96Assaf may has more information.

You are right, it works on the **develop branch**, and as you already mentioned, the problem is specifically related to whether the **tree has a sensor or not**.

This means that the PR is now redundant and can actually be deleted.

However, we still need to find a solution for the issue with trees without sensors, but I'm not sure if this issue is so critical right now. 

We can fix it on the frontend side:
In this case, if a user selects only one tree while creating a tree cluster and the tree does not have a sensor, we can simply display a toast notification informing them that the watering status will be **unknown**.

choffmann:
I suggest checking the watering status of the tree every time you touch it to see if it can be so, a kind of plausibility check, easy peasy. But we can also create an issue for that and close this pr

doriengr:
I am not quite getting what now the problem is expect that the cluster is still not correctly updated in some cases?
Did someone check when it does not work? 

Frontend-wise we should not implement a toast — in my point of view this is a backend problem.

PR #443, opened 2025-02-28
Refactor/cleanup code and delete images

doriengr:


internal/storage/auth/keycloak/keycloak.go
```diff
@@ -66,8 +66,6 @@ func NewKeycloakRepository(cfg *config.IdentityAuthConfig) storage.AuthRepositor
 //	if err != nil {
 //	    log.Fatalf("Failed to authenticate: %v", err)
 //	}
-//
-//	fmt.Printf("Access Token: %s\n", token.AccessToken)
```
choffmann, line 70: That is part of the `Example` section in the doc comment

internal/storage/postgres/tree/get.go
```diff
@@ -241,12 +230,8 @@ func (r *TreeRepository) getTreeClusterByTreeID(ctx context.Context, treeID int3
 	return tc, nil
 }
 
-// Map sensor, images and tree cluster entity to domain tree
+// Map sensor and tree cluster entity to domain tree
 func (r *TreeRepository) mapFields(ctx context.Context, t *entities.Tree) error {
```
choffmann, line 234: You don't touch it but why does `mapTreeCluster` return an error if we ignore it anyway?

pkg/client/model_tree.go
choffmann, line 1: Did you remove this line by hand or regenerate it with `make generate/client` because I get more changes when I run this command. It's not image specific, but there are some changes for the client, so can you run `make generate/client` and commit those changes here?

choffmann (changes requested):

choffmann (approved):

PR #446, opened 2025-02-28
Refactor/implement some various todos

### LINKED ISSUE #444: Go through code and implement // TODOS

doriengr:
refs #444 

## Stuff that was updated
- moved filtering of users by role to storage layer
- deleted todos that were already implemented or not needed anymore
- handle error on mapping sensor data
- asserted sensor data in sensor handler

### DIFF
--- a/internal/storage/auth/keycloak/user.go
+++ b/internal/storage/auth/keycloak/user.go
@@ -12,6 +12,7 @@ import (
 	"github.com/green-ecolution/green-ecolution-backend/internal/config"
 	"github.com/green-ecolution/green-ecolution-backend/internal/entities"
 	"github.com/green-ecolution/green-ecolution-backend/internal/logger"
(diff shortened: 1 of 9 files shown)


internal/storage/auth/keycloak/user.go
```diff
@@ -121,6 +122,27 @@ func (r *UserRepository) GetAll(ctx context.Context) ([]*entities.User, error) {
...
+	for _, user := range users {
+		for _, userRole := range user.Roles {
+			if userRole == role {
+				filteredUsers = append(filteredUsers, user)
+				break
+			}
+		}
+	}
+
+	return filteredUsers, nil
```
choffmann, line 143: You could use some standard iterator functions such as filter, which are defined in the utils package (`utils.Filter`), to reduce the ugly for loops, but this is up to you :man_shrugging:

choffmann (approved):
:+1:

PR #452, opened 2025-03-03
feat: map sensor to tree when it sends data from ttn

### LINKED ISSUE #445: Map sensor to tree

Mohammad96Assaf:
close #445

### DIFF
--- a/internal/service/domain/sensor/mqtt.go
+++ b/internal/service/domain/sensor/mqtt.go
@@ -30,14 +30,6 @@ func (s *SensorService) HandleMessage(ctx context.Context, payload *domain.MqttP
 		}
 	}
 
-	if err != nil {
-		var entityNotFoundErr storage.ErrEntityNotFound
-		if !errors.As(err, &entityNotFoundErr) {
-			log.Error("failed to get sensor by id", "error", err)
-			return nil, err
-		}
-	}
-
 	if sensor != nil {
 		updatedSensor, err := s.updateSensorCoordsAndStatus(ctx, payload, sensor)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


Mohammad96Assaf:
For now, the sensor will be mapped to the nearest tree within a 3-meter distance. When the sensor sends data, it will either be updated if it exists or created if it cannot be found.

This means the sensor is always stored in the database. However, the question is: what should we do if no nearby tree can be found for the sensor from TTN? Should we delete it?

I’d like to discuss this point. Please leave a comment under this PR if you have a better idea for this approach or if we should keep the implementation as it is now.

doriengr:
> For now, the sensor will be mapped to the nearest tree within a 3-meter distance. When the sensor sends data, it will either be updated if it exists or created if it cannot be found.
> 
> This means the sensor is always stored in the database. However, the question is: what should we do if no nearby tree can be found for the sensor from TTN? To me, it doesn’t make sense to have a sensor without a tree.
> 
> I’d like to discuss this point. Please leave a comment under this PR if you have a better idea for this approach or if we should keep the implementation as it is now.

I would still create the sensor if no tree can be found. We also pitched it this way already to the tbz and smarte grenzregion. 
In the frontend & backend there is a possibility to link a sensor to a tree afterwards. So if for example the GPS data is a little bit off or something is not working while mapping the sensor to a tree, the users still have a possibility to manually create a connection.

Mohammad96Assaf:
@doriengr @choffmann 
Okey, that sounds good. If you don't have any other suggestions, please approve the PR

choffmann (approved):
Tested it whith a simulated mqtt call, seems to work :+1:

PR #457, opened 2025-03-05
feat: add filter to tree

### LINKED ISSUE #84: Implement filter functions in trees
Implement the ability to get trees based on a filter.

Possible filters are:
- Watering condition (good, moderate, bad)
- Region
- Planting Year

Change the http handler to accept filter option. Maybe over query parameter `GET /tree?status=good&region=altstadt&year_start=2022&year_end=2024`

Mohammad96Assaf:
close #84

### DIFF
--- a/internal/entities/tree.go
+++ b/internal/entities/tree.go
@@ -49,8 +49,8 @@ type TreeUpdate struct {
 }
 
 type TreeQuery struct {
-	WateringStatus []WateringStatus `query:"watering_statuses"`
-	HasCluster     *bool            `query:"has_cluster"`
-	PlantingYears  []int32          `query:"planting_years"`
+	WateringStatuses []WateringStatus `query:"watering_statuses"`
+	HasCluster       *bool            `query:"has_cluster"`
+	PlantingYears    []int32          `query:"planting_years"`
 	Query
 }
--- a/internal/entities/tree.go
+++ b/internal/entities/tree.go
@@ -50,7 +50,7 @@ type TreeUpdate struct {
 
 type TreeQuery struct {
 	WateringStatus []WateringStatus `query:"status"`
-	IsInCluster    *bool            `query:"in_cluster"`
+	HasCluster     *bool            `query:"hasCluster"`
 	Years          []int32          `query:"years"`
 	Query
 }
--- a/internal/entities/tree.go
+++ b/internal/entities/tree.go
@@ -49,8 +49,8 @@ type TreeUpdate struct {
 }
 
 type TreeQuery struct {
-	WateringStatus []WateringStatus `query:"status"`
-	HasCluster     *bool            `query:"hasCluster"`
-	Years          []int32          `query:"years"`
+	WateringStatus []WateringStatus `query:"watering_statuses"`
+	HasCluster     *bool            `query:"has_cluster"`
+	PlantingYears  []int32          `query:"planting_years"`
 	Query
 }
--- a/internal/entities/tree.go
+++ b/internal/entities/tree.go
@@ -47,3 +47,11 @@ type TreeUpdate struct {
 	Provider       string
 	AdditionalInfo map[string]interface{}
 }
+
+type TreeQuery struct {
+	WateringStatus []WateringStatus `query:"status"`
+	Region         []string         `query:"region"`
+	YearStart      int32            `query:"year_start"`
+	YearEnd        int32            `query:"year_end"`
+	Query
+}
--- a/internal/entities/tree.go
+++ b/internal/entities/tree.go
@@ -51,7 +51,6 @@ type TreeUpdate struct {
 type TreeQuery struct {
 	WateringStatus []WateringStatus `query:"status"`
 	Region         []string         `query:"region"`
-	YearStart      int32            `query:"year_start"`
-	YearEnd        int32            `query:"year_end"`
+	Years          []int32          `query:"years"`
 	Query
 }
--- a/internal/entities/tree.go
+++ b/internal/entities/tree.go
@@ -50,7 +50,7 @@ type TreeUpdate struct {
 
 type TreeQuery struct {
 	WateringStatus []WateringStatus `query:"status"`
-	Region         []string         `query:"region"`
+	IsInCluster    *bool            `query:"in_cluster"`
 	Years          []int32          `query:"years"`
 	Query
 }
--- a/internal/server/http/handler/v1/tree/handler.go
+++ b/internal/server/http/handler/v1/tree/handler.go
@@ -33,8 +33,9 @@ var (
 //	@Param			page		query	int		false	"Page"
 //	@Param			limit		query	int		false	"Limit"
 //	@Param			provider	query	string	false	"Provider"
-// 	@Param			status		query	string	false	"watering status (good, moderate, bad)"
-// 	@Param			years		query	[]int	false	"years"
+// @Param			status		query	string	false	"watering status (good, moderate, bad)"
+// @Param			years		query	[]int	false	"years"
+// @Param			in_cluster	query	bool	false	"is in cluster"
 //	@Security		Keycloak
 func GetAllTrees(svc service.TreeService) fiber.Handler {
 	return func(c *fiber.Ctx) error {
--- a/internal/server/http/handler/v1/tree/handler.go
+++ b/internal/server/http/handler/v1/tree/handler.go
@@ -63,22 +63,22 @@ func GetAllTrees(svc service.TreeService) fiber.Handler {
 	}
 }
 
-func fillTreeQueryParams(c *fiber.Ctx) (*domain.TreeQuery, error) {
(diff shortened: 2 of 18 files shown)


internal/entities/tree.go
```diff
@@ -47,3 +47,11 @@ type TreeUpdate struct {
 	Provider       string
 	AdditionalInfo map[string]interface{}
 }
+
+type TreeQuery struct {
+	WateringStatus []WateringStatus `query:"status"`
```
doriengr, line 52: image

I am not sure if those values are correct. 

Missing is the filter if a tree is in a tree_cluster or not.
And I do not know if `year_start` and `year_end` is needed. I think `years` is better since a user can select multiple years: for example 2022, 2023 and 2025 but not 2024.

internal/entities/tree.go
```diff
@@ -47,3 +47,11 @@ type TreeUpdate struct {
...
+	Region         []string         `query:"region"`
```
doriengr, line 53: Trees do not have regions

internal/entities/tree.go
Mohammad96Assaf: Please have a look at the comment from @choffmann in #84. 
Should I change it to allow users to select multiple years instead of "year_start and year_end"?

internal/entities/tree.go
Mohammad96Assaf: Please have a look at the comment from @choffmann in #84. 

In the query I used the region from Cluster.

Mohammad96Assaf: should I remove "region" option ?

internal/entities/tree.go
doriengr: The description of the issue was not updated since Sep 2024, so its sadly not up-to-date anymore.
Since we do not have the time to make big changes in the frontend for the filters, the years should be a multiselect for the years 👍

internal/entities/tree.go
doriengr: Yes please :)

internal/entities/tree.go
```diff
@@ -47,3 +47,10 @@ type TreeUpdate struct {
...
```
doriengr, line 53: I am not sure if you already changed everything but the region is still in the struct. It has to be deleted. 

Also there is no filter option if a tree is in a cluster or not. This is also still missing. Please check out the current filter in the frontend to see the needed behaviour. 👍

internal/entities/tree.go
```diff
@@ -47,3 +47,10 @@ type TreeUpdate struct {
 	Provider       string
 	AdditionalInfo map[string]interface{}
 }
+
+type TreeQuery struct {
+	WateringStatus []WateringStatus `query:"status"`
+	HasCluster     *bool            `query:"hasCluster"`
```
doriengr, line 53: `query:has_cluster`

doriengr, line 53: `query:has_cluster`

internal/server/http/handler/v1/tree/handler.go
```diff
@@ -33,17 +33,20 @@ var (
 // @Param			page		query	string	false	"Page"
 // @Param			limit		query	string	false	"Limit"
 // @Param			provider	query	string	false	"Provider"
+// @Param			status		query	string	false	"watering status (good, moderate, bad)"
+// @Param			years		query	[]int	false	"years"
+// @Param			hasCluster	query	bool	false	"has cluster"
```
doriengr, line 38: `// @Param has_cluster query bool false "has cluster"`

internal/server/http/handler/v1/tree/handler_test.go
```diff
@@ -26,7 +26,7 @@ func TestGetAllTrees(t *testing.T) {
 		app.Get("/v1/tree", tree.GetAllTrees(mockTreeService))
 		mockTreeService.EXPECT().GetAll(
 			mock.Anything,
-			entities.Query{},
+			&entities.TreeQuery{},
```
doriengr, line 29: No test with filter options?

internal/service/domain/tree/tree_test.go
```diff
@@ -30,10 +30,10 @@ func TestTreeService_GetAll(t *testing.T) {
 		svc := tree.NewTreeService(treeRepo, sensorRepo, clusterRepo, eventManager)
 
 		expectedTrees := TestTreesList
-		treeRepo.EXPECT().GetAll(ctx, entities.Query{}).Return(expectedTrees, int64(len(expectedTrees)), nil)
+		treeRepo.EXPECT().GetAll(ctx, &entities.TreeQuery{}).Return(expectedTrees, int64(len(expectedTrees)), nil)
```
doriengr, line 33: No tests with filter options?

internal/storage/postgres/tree/get_test.go
```diff
@@ -19,7 +19,7 @@ func TestTreeRepository_GetAll(t *testing.T) {
 		ctx = context.WithValue(ctx, "limit", int32(-1))
 
 		// when
```
doriengr, line 21: There are tests missing to test the filter options

doriengr (changes requested):

internal/service/domain/tree/tree.go
```diff
@@ -39,7 +39,7 @@ func NewTreeService(
 	}
 }
 
-func (s *TreeService) GetAll(ctx context.Context, query entities.Query) ([]*entities.Tree, int64, error) {
+func (s *TreeService) GetAll(ctx context.Context, query *entities.TreeQuery) ([]*entities.Tree, int64, error) {
```
doriengr, line 42: Why is the Query in the treecluster service just `entities.TreeClusterQuery`

internal/entities/tree.go
```diff
@@ -47,3 +47,10 @@ type TreeUpdate struct {
...
+	Years          []int32          `query:"years"`
```
doriengr, line 54: Sorry, but we should call this `planting_years` since it is also called like this in the frontend

internal/entities/tree.go
```diff
@@ -47,3 +47,10 @@ type TreeUpdate struct {
 	Provider       string
 	AdditionalInfo map[string]interface{}
 }
+
+type TreeQuery struct {
+	WateringStatus []WateringStatus `query:"status"`
```
doriengr, line 52: `WateringStatuses []WateringStatus query:"watering_statuses"`

Because it is an array

internal/service/domain/tree/tree.go
Mohammad96Assaf: The linter requires addiing  a pointer (*) to `treeQuery` because it has two lists and one pointer filed inside it. I'm not exactly sure why, but without the pointer (*), the linter shows an error. So we have to use a pointer (*) to make the linter happy.

internal/entities/tree.go
```diff
@@ -47,3 +47,10 @@ type TreeUpdate struct {
 	Provider       string
 	AdditionalInfo map[string]interface{}
 }
+
+type TreeQuery struct {
+	WateringStatus []WateringStatus `query:"watering_statuses"`
```
doriengr, line 52: `WateringStatuses []WateringStatus`

internal/server/http/handler/v1/tree/handler.go
```diff
@@ -33,17 +33,20 @@ var (
 //	@Param			page		query	int		false	"Page"
 //	@Param			limit		query	int		false	"Limit"
 //	@Param			provider	query	string	false	"Provider"
-//	@Security		Keycloak
+// 	@Param			watering_statuses	query	string	false	"watering status (good, moderate, bad)"
```
doriengr, line 36: This should also be an array of strings :) 

`// @Param			watering_statuses	query	[]string	false	"Watering statuses"`

internal/service/domain/tree/tree.go
doriengr: Maybe I am mistaken but in the treeQuery there are lists but none of them has a pointer? So is this still needed?

doriengr: I am just wondering because I do not like it that this Query needs a pointer and the one for treecluster does not. We should unify this in my eyes and I am hoping that we do not need the pointer here anymore.

doriengr (changes requested):

internal/service/domain/tree/tree.go
Mohammad96Assaf: I understand that unifying them would be better, but the query structs themselves are different. The `TreeQuery` struct isn't the same as the `TreeClusterQuery` struct, so sometimes their usage will differ.

The `TreeQuery` struct has slices and pointer fields `(like *bool)`. Because of these fields, the linter requires using a pointer `(*)`.  If you omit the pointer `(*)`, the linter will show an error.

You can remove the pointer `(*)` yourself and see how the linter responds :)

doriengr (approved):

PR #458, opened 2025-03-11
feat: implement demo

choffmann:
add ability to:
- disable auth
- disable routing services
- disable s3
- disable mqtt


doriengr (approved):
Did not test it, but it looks good. 

Just one question:
Maybe we should add a comment in the files of the dummy stuff just to clarify for what those files are needed?

PR #461, opened 2025-03-14
refactor: unify param types of pagination values in swagger

doriengr:
ref #393 

limit and page should be an `int` in every endpoint, 
and filtering option in the tree_cluster should be arrays


choffmann (approved):
Nice :fire:

PR #462, opened 2025-03-16
feat: add docker compose file to run application locally

choffmann:
Add a docker compose file to run a demo of the application locally. The mqtt, auth, routing and s3 services will be disabled by default. This file is mentioned in the documentation. 

The compose file will spin up a postgres db and insert the migrations. Then it will start the frontend and backend with traefik as the reverse proxy.


doriengr (approved):
🔥

PR #463, opened 2025-03-17
Refactor/cleanup tree cluster query in handler

doriengr:


internal/server/http/handler/v1/tree/handler.go
```diff
@@ -63,24 +63,6 @@ func GetAllTrees(svc service.TreeService) fiber.Handler {
...
 
-func fillTreeQueryParams(c *fiber.Ctx) (domain.TreeQuery, error) {
-	var filter domain.TreeQuery
-
-	if err := c.QueryParser(&filter); err != nil {
-		return domain.TreeQuery{}, err
-	}
-
-	if c.Query("status") != "" {
-		wateringStatuses, err := domain.ParseWateringStatus(c.Query("status"))
```
Mohammad96Assaf, line 74: I think the function `ParseWateringStatus(status string) ([]WateringStatus, error)` in `entities/watering_status.go` can also be removed as it is redundant now.

Mohammad96Assaf (changes requested):

internal/server/http/handler/v1/tree/handler.go
Mohammad96Assaf: > I think the function `ParseWateringStatus(status string) ([]WateringStatus, error)` in `entities/watering_status.go` can also be removed as it is redundant now.

inside the tests, where the function is used, we can directly create a `WateringStatus` slice: ` []entities.WateringStatus{}`

Mohammad96Assaf (approved):

Issue #466, opened 2025-03-22
Unable to remove tree from treecluster


xk0niR:
If you try to remove a tree from a treecluster nothing will change.
GIF as example:

![Image](https://github.com/user-attachments/assets/b8b0ef49-7052-4ae3-8587-d334494a73fa)

PR #467, opened 2025-03-23
Bugfix/tree can not be removed from tree cluster

doriengr:
Close https://github.com/green-ecolution/green-ecolution-backend/issues/466


xk0niR (approved):
Works now as intended :+1:

PR #470, opened 2025-03-24
fix: only one user needs required driving licences

### LINKED ISSUE #469: Creating watering plan with min. 1 person with correct driving license
When I create a watering plan with two persons: 1 has all needed driving licences and the other does not — I still get an error. That should not happen since one person can drive the vehicle.

Image

doriengr:
Close #469


xk0niR (approved):
How I tested:
- Choose vehicle which requires driving license C
- choose Person X which only has B => Cant create ("Mindestens ein Mitarbeitender muss die erforderliche Führerscheinklasse besitzen.") OK
- choose Person X and Y and Y has C =>  Can create OK

:1st_place_medal:

PR #476, opened 2025-03-26
feat: add provider and additional info to in list responses

choffmann:


choffmann:
I will merge it because I need it for the pipeline at https://github.com/green-ecolution/green-ecolution-frontend/pull/478

PR #477, opened 2025-03-29
feat: recalculate cluster watering status

### LINKED ISSUE #473: Watering status is not updated on tree cluster create
How to reproduce: 
1. Create a treecluster where one tree has a sensor
2. Create another treecluster with the the same tree
3. Second tree cluster has correct watering status but the first treecluster (without the tree and its sensor) is not updated correctly:

Image

This only happens if you create a new tree cluster, on update everything works correctly.

choffmann:
Close #473


choffmann:
> I have tested this successfully as described in #473 . build_and_test is failing as seeing above. Can you check this @choffmann ? 

@xk0niR done!

choffmann:
> EDIT: The issue is already mentioned in #472

See #478

xk0niR (approved):
:100:

PR #478, opened 2025-03-29
fix: set region to nil if trees are empty

### LINKED ISSUE #258: Tree cluster without trees still has a region
It's not a bad bug since it does not break anything. Low priority

Image

Reproduce:
1. Create a tree cluster "A" with a tree
2. Update another tree cluster and select the linked tree of cluster "A" to link it to the current cluster
3. tree cluster "A" still has a region set — but it should be empty right?

### LINKED ISSUE #472: Tree cluster without trees still has a region
It's not a bad bug since it does not break anything. Low priority

Image

Reproduce:
1. Create a tree cluster "A" with a tree
2. Update another tree cluster and select the linked tree of cluster "A" to link it to the current cluster
3. tree cluster "A" still has a region set — but it should be empty right?

choffmann:
Close green-ecolution/green-ecolution#258

### DIFF
--- a/internal/service/domain/treecluster/treecluster.go
+++ b/internal/service/domain/treecluster/treecluster.go
@@ -307,6 +307,8 @@ func (s *TreeClusterService) updateTreeClusterPosition(ctx context.Context, id i
 					slog.Group("new_position", "latitude", lat, "longitude", long),
 				)
 			}
+		} else {
+			tc.Region = nil
 		}
 
 		return true, nil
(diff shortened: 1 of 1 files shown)


xk0niR (changes requested):
I still get this error.

After applying seed data:
- Create treecluster A with tree 1030 (only yellow one)
- Create treecluster B with tree 1030 (only yellow one), stealing from A.

=> A still shows Jürgensby as region without any tree.

PR #506, opened 2025-07-11
feat: embed frontend in binary

choffmann:
To build single binary like in the new root [flake.nix](https://github.com/green-ecolution/green-ecolution/blob/develop/flake.nix)


akatranlp (approved):
👍

PR #507, opened 2025-07-12
feat: check in generated code

choffmann:
CI tested with client package not updated:
grafik
https://github.com/green-ecolution/backend/actions/runs/16237954870/job/45850536359


akatranlp (approved):
it works, so it must be right!

PR #513, opened 2025-07-12
feat: override postgres date types to time.Time

akatranlp:
In this PR all, pg_types are replaced with go types. `timestamp`, `timestamptz` and `date` with `time.Time` and `uuid` with `github.com/google/uuid.UUID`


choffmann (approved):
🔥

